PR #1, opened 2019-09-23
V1

lakenyon:
Initial 1.0 version.


Playgrounds/Codables.swift
```diff
@@ -0,0 +1,125 @@
+import UIKit
```
darinkrauss: Can we drop the `Playgrounds` directory and `TidepoolKit.playground`? No other Loop related repo has playgrounds committed.

TidepoolKit/Types/TPDeviceData/TPDataCbg.swift
```diff
@@ -0,0 +1,89 @@
...
+
+import Foundation
+import HealthKit
+
+public enum TPCbgUnit: String, Encodable {
+    case milligramsPerDeciliter = "mg/dL"
+    case millimolesPerLiter = "mmol/L"
+    
+    // service syntax check
+    func inRange(_ value: Double) -> Bool {
```
darinkrauss: Let's drop the range validation.

TidepoolKit/Types/TPDeviceData/TPDataCbg.swift
```diff
@@ -0,0 +1,89 @@
...
+    // MARK: - Type specific data
+    //
+    public let value: Double
+    public let units: TPCbgUnit
+
+    /// Only values acceptable to the Tidepool service are allowed in creating a TPDataCbg item.
+    public init(time: Date, value: Double, units: TPCbgUnit) {
+        self.value = value
+        self.units = units
+        if !self.units.inRange(self.value) {
```
darinkrauss: Drop the range validation.

TidepoolKit/Types/TPDeviceData/TPDataCbg.swift
```diff
@@ -0,0 +1,89 @@
...
+            LogError("TPDataCbg init: value \(value) \(units.rawValue) out of range!")
+        }
+        // TPSampleData fields
+        super.init(.cbg, time: time)
+    }
+    
+    //
+    // MARK: - RawRepresentable
```
darinkrauss, line 29: I think we can drop a lot of the `MARK`s. In general, we've only used those to separate out different chunks of logic in large files.

TidepoolKit/Types/TPDeviceData/TPDataCbg.swift
```diff
@@ -0,0 +1,89 @@
...
+            return nil
+        }
+        guard let units = TPCbgUnit(rawValue: unitsString) else {
+            LogError("TPDataCbg:init(rawValue) invalid units found!")
+            return nil
+        }
+        self.value = value.doubleValue
+        self.units = units
+        
+        // base properties in superclass...
```
darinkrauss: Not sure that these types of comments are necessary. There aren't a lot of inline comments in Loop. Pretty much just limited to areas of code where the logic is complicated or the code is unexpected, not where the code is expected.

TidepoolKit/Types/TPDeviceData/TPDataCbg.swift
```diff
@@ -0,0 +1,89 @@
...
+        super.init(rawValue: rawValue)
+    }
+    
+    override public var rawValue: RawValue {
+        // start with common data
+        var dict = super.rawValue
```
darinkrauss: Generally, we've used `rawValue` local variable rather than `dict` in this situation.

TidepoolKit/Types/TPDeviceData/TPDataCbg.swift
```diff
@@ -0,0 +1,89 @@
...
+//  TidepoolKit
+//
+//  Created by Larry Kenyon on 8/23/19.
+//  Copyright © 2019 Tidepool Project. All rights reserved.
+//
+
+import Foundation
+import HealthKit
+
+public enum TPCbgUnit: String, Encodable {
```
darinkrauss: This isn't limited to `cbg`, but rather common to all blood glucose values (regardless of source, `cbg`, `bgm`, or otherwise).

TidepoolKit/Networking/APIAccessUsers.swift
```diff
@@ -0,0 +1,68 @@
...
+//
+
+import Foundation
+
+public class APIAccessUsers: TPFetchable {
+    
+    public let users: [TPUser]
+        
+    //
+    // MARK: - methods private to framework!
```
darinkrauss: Probably not needed.

TidepoolKit/Networking/APIAccessUsers.swift
```diff
@@ -0,0 +1,68 @@
...
+    init(_ users: [TPUser]) {
+        self.users = users
+    }
+
+    //
+    // MARK: - TPFetchable protocol conformance methods
+    //
+    
+    class func urlExtension(forUser userId: String) -> String {
+        let urlExtension = "/access/groups/" + userId
```
darinkrauss: Consider `return "/access/groups/\(userId)"`.

TidepoolKit/Networking/APIAccessUsers.swift
```diff
@@ -0,0 +1,68 @@
...
+
+public class APIAccessUsers: TPFetchable {
+    
+    public let users: [TPUser]
+        
+    //
+    // MARK: - methods private to framework!
+    //
+
+    init(_ users: [TPUser]) {
```
darinkrauss: I would even go so far as to make it `private` since the only time it will be constructed is from the `fromJsonData` below.

TidepoolKit/Networking/APIAccessUsers.swift
```diff
@@ -0,0 +1,68 @@
...
+        self.users = users
+    }
+
+    //
+    // MARK: - TPFetchable protocol conformance methods
+    //
+    
+    class func urlExtension(forUser userId: String) -> String {
```
darinkrauss: Initially confused by the `urlExtension` name. Consider `urlPath` as a more accurate name.

TidepoolKit/Networking/APIAccessUsers.swift
```diff
@@ -0,0 +1,68 @@
...
+                users.append(user)
+                print("adding user: \(user)")
+            }
+        }
+        return APIAccessUsers(users)
+    }
+}
+
+
+/*
```
darinkrauss: I would avoid all of the JSON examples. They are likely to just go stale over time and cause confusion. The API spec in Stoplight is the definitive API documentation.

TidepoolKit/Networking/APIAccessUsers.swift
```diff
@@ -0,0 +1,68 @@
...
+    class func fromJsonData(_ data: Data) -> TPFetchable? {
+        guard let jsonDict = try? JSONSerialization.jsonObject(with: data, options: []) as? NSDictionary else {
+            return nil
+        }
+        var users: [TPUser] = []
+        for key in jsonDict.keyEnumerator() {
+            if let keyStr = key as? String {
+                let user = TPUser(keyStr)
+                users.append(user)
+                print("adding user: \(user)")
```
darinkrauss: Probably should drop most of the `print` statements.

TidepoolKit/Networking/APIAccessUsers.swift
```diff
@@ -0,0 +1,68 @@
...
+        return urlExtension
+    }
+    
+    class func fromJsonData(_ data: Data) -> TPFetchable? {
+        guard let jsonDict = try? JSONSerialization.jsonObject(with: data, options: []) as? NSDictionary else {
+            return nil
+        }
+        var users: [TPUser] = []
+        for key in jsonDict.keyEnumerator() {
+            if let keyStr = key as? String {
```
darinkrauss: It took me a while to figure this out, but `keyStr` is really `userId`, isn't it? If so, I'd user `userId` here to make it more clear.

TidepoolKit/Networking/APIAccessUsers.swift
```diff
@@ -0,0 +1,68 @@
...
+                let user = TPUser(keyStr)
```
darinkrauss: So we are only capturing the `userId` here? Not the rest of the structure?

TidepoolKit/Networking/APIConnect.swift
```diff
@@ -0,0 +1,994 @@
...
+//  TidepoolKit
+//
+//  Created by Larry Kenyon on 8/23/19.
+//  Copyright © 2019 Tidepool Project. All rights reserved.
+//
+
+import Foundation
+
+typealias NetworkRequestCompletionHandler = (Data?, URLResponse?, Error?) -> Void
+protocol TidepoolNetworkInterface {
```
darinkrauss, line 13: [nit] Blank line between `NetworkRequestCompletionHandler` and `TidepoolNetworkInterface`?

TidepoolKit/Networking/APIConnect.swift
```diff
@@ -0,0 +1,994 @@
...
+//  APIConnector.swift
+//  TidepoolKit
+//
+//  Created by Larry Kenyon on 8/23/19.
+//  Copyright © 2019 Tidepool Project. All rights reserved.
+//
+
+import Foundation
+
+typealias NetworkRequestCompletionHandler = (Data?, URLResponse?, Error?) -> Void
```
darinkrauss, line 11: I would make a `struct` that joins the `Data` and `URLResponse` objects. Maybe call it `ResponseData`. Then change the handler to `Result<ResponseData, Error>. Then, ditch the `typealias` because it can be easily represented as `(Result<ResponseData, Error>) -> Void`.

This also makes it clear that there will either be a `ResponseData` with `Data` and `URLResponse` *OR* and `Error`.

TidepoolKit/Networking/APIConnect.swift
```diff
@@ -0,0 +1,994 @@
...
+/// - Provides online/offline status.
+/// - Get/put user data of various types (cbg, carb, etc.)
+class APIConnector {
+    
+    // non-nil when "loggedIn", nil when "loggedOut"
+    var session: TPSession?
+    var apiQueue: DispatchQueue
+    
+    // Base URL for API calls, set during initialization
+    var baseUrlString: String? {
```
darinkrauss: Consider making this an entirely computed property backed by the `baseURL` property?

TidepoolKit/Networking/APIConnect.swift
```diff
@@ -0,0 +1,994 @@
...
+        didSet {
+            if let urlStr = baseUrlString {
+                self.baseUrl = URL(string: urlStr)!
```
darinkrauss: What if this fails?

TidepoolKit/Networking/APIConnect.swift
```diff
@@ -0,0 +1,994 @@
...
+            } else {
+                self.baseUrl = nil
+            }
+        }
+    }
+    var baseUrl: URL?
```
darinkrauss: Actually, why is there even a `baseUrl` in the first place? I don't see it being used.

TidepoolKit/Networking/APIConnect.swift
```diff
@@ -0,0 +1,994 @@
...
+
+    /// Reachability object, valid during lifetime of this APIConnector, and convenience function that uses this
+    var reachability: ReachabilitySource?
+    var networkRequestHandler: TidepoolNetworkInterface
+    
+    init(queue: DispatchQueue) {
+        LogInfo("")
```
darinkrauss: ???

TidepoolKit/Networking/APIConnect.swift
```diff
@@ -0,0 +1,994 @@
...
+    
+    // MARK: - Constants
+    
+    private let kSessionTokenHeaderId = "X-Tidepool-Session-Token"
+    private let kSessionTokenResponseId = "x-tidepool-session-token"
+    
+    // Dictionary of servers and their base URLs
+    private func serverUrl(_ server: TidepoolServer) -> String {
+        switch server {
+        case .development: return "https://dev-api.tidepool.org"
```
darinkrauss: [nit] Separate lines?

TidepoolKit/Networking/APIConnect.swift
```diff
@@ -0,0 +1,994 @@
...
+        case .staging: return "https://stg-api.tidepool.org"
+        case .integration: return "https://int-api.tidepool.org"
+        case .production: return "https://api.tidepool.org"
+        }
+    }
+    private let kDefaultServer: TidepoolServer = .staging
```
darinkrauss: I would add the default to the `TidepoolServer` enum itself as a computed property so it could be used like `TidepoolService.default`.

TidepoolKit/Networking/APIConnect.swift
```diff
@@ -0,0 +1,994 @@
...
+        }
+        self.session = session
+        self.baseUrlString = serverUrl(session.server)
+        LogInfo("Logged in with existing TPSession!")
+        NotificationCenter.default.post(name: TidepoolLogInChangedNotification, object:self)
+        return Result.success(session)
+    }
+    
+    func clearSession() {
+        LogVerbose("")
```
darinkrauss: ???

darinkrauss (changes requested):
Partial review.

TidepoolKit/Networking/APIConnect.swift
lakenyon: Removed, it was a remnant of a previous version...

TidepoolKit/Networking/APIConnect.swift
lakenyon: LogInfo takes a string; with a nil string, the function, file, etc are still logged.

TidepoolKit/Networking/APIConnect.swift
lakenyon: Changed, but this is going away with the Kubernetes support.

TidepoolKit/Networking/APIConnect.swift
lakenyon: Again, will be changing with the Kubernetes support.

TidepoolKit/Networking/APIConnect.swift
lakenyon: Same as above.

TidepoolKit/Types/TPDeviceData/TPDataCbg.swift
lakenyon: I've now mostly removed these except for a couple large files where they do help show overall organization and help to get the reader to a specific area, and where they are used to set off protocol conformance (mainly for rawRepresentable, where that can't be done via an extension of a class). Also removed the extra // lines around the MARK statements to match usage in Loop.

TidepoolKit/Types/TPDeviceData/TPDataCbg.swift
lakenyon: Removed a number of these, except where they help me remember what's going on ;-)

Playgrounds/Codables.swift
lakenyon: Done.

TidepoolKit/Types/TPDeviceData/TPDataCbg.swift
lakenyon: Done. Removed all validations on object inits except from raw values.

TidepoolKit/Types/TPDeviceData/TPDataCbg.swift
lakenyon: Done.

TidepoolKit/Types/TPDeviceData/TPDataCbg.swift
lakenyon: Done.

TidepoolKit/Networking/APIAccessUsers.swift
lakenyon: Removed.

TidepoolKit/Types/TPDeviceData/TPDataCbg.swift
lakenyon: Moved to separate file.

TidepoolKit/Networking/APIAccessUsers.swift
lakenyon: Why?

TidepoolKit/Networking/APIAccessUsers.swift
lakenyon: fromJsonData is now in an extension; this is not public, but needs to be at least "internal" for it to be accessible in fromJsonData.

TidepoolKit/Networking/APIAccessUsers.swift
lakenyon: Renamed.

TidepoolKit/Networking/APIAccessUsers.swift
lakenyon: Removed.

TidepoolKit/Networking/APIAccessUsers.swift
lakenyon: Removed and/or changed to Log statements.

TidepoolKit/Networking/APIAccessUsers.swift
lakenyon: Yes, changed!

TidepoolKit/Networking/APIAccessUsers.swift
lakenyon: Yes, that is all that has been used in Tidepool Mobile to date. If Loop or other clients need more, we should prioritize.

TidepoolKit/Networking/APIConnect.swift
lakenyon: Added.

TidepoolKit/Networking/APIConnect.swift
lakenyon: This is a common trio of values coming from URLSession calls, the code dispatches these at a very low level and quickly incorporates them into a struct along these lines. Not sure it's worth the effort to change this.

TidepoolKit/Networking/APIConnect.swift
lakenyon: Now just a string, used in 2 places to create a URL.

TidepoolKit/Networking/APIConnect.swift
lakenyon: Should only be due to a coding error - ! causing a crash should be fine. Anyway, now just a string since baseUrl wasn't being used anymore.

TidepoolKitUITests/TidepoolKitUITests.swift
```diff
@@ -0,0 +1,34 @@
+//
```
darinkrauss: Remove since there are no tests in this file?

TidepoolKitUI/TidepoolKitUI.swift
```diff
@@ -0,0 +1,34 @@
...
+//  TidepoolKit
+//
+//  Created by Larry Kenyon on 8/23/19.
+//  Copyright © 2019 Tidepool Project. All rights reserved.
+//
+
+import Foundation
+import TidepoolKit
+
+
```
darinkrauss: [nit] Remove extra blank line?

TidepoolKitUI/TidepoolKitUI.swift
```diff
@@ -0,0 +1,34 @@
...
+public class TidepoolKitUI {
+    
+    private var tpKit: TidepoolKit
```
darinkrauss: [nit] Rename `tpKit` -> `tidepoolKit` throughout.

TidepoolKitUI/TidepoolKitUI.swift
```diff
@@ -0,0 +1,34 @@
...
+    
+    public init(tpKit: TidepoolKit, logger: TPKitLogging? = nil) {
+        clientLogger = logger
+        self.tpKit = tpKit
+    }
+    
+    public func logInViewController(loginSignupDelegate: LoginSignupDelegate? = nil, defaultServerHost: String?) -> UIViewController {
+        let loginViewController = UIStoryboard(name: "LoginSignup", bundle: Bundle(for: LoginViewController.self)).instantiateInitialViewController() as! LoginViewController
+        loginViewController.loginSignupDelegate = loginSignupDelegate
+        loginViewController.tpKit = self.tpKit
```
darinkrauss: [nit] Don't need `self.` here.

TidepoolKitUI/TidepoolKitUI.swift
```diff
@@ -0,0 +1,34 @@
...
+        loginViewController.serverHost = defaultServerHost
+        return loginViewController
+    }
+
+}
+
+// global logging protocol, optional...
+var clientLogger: TPKitLogging?
```
darinkrauss: [nit] Its not really clear what `clientLogger` means in this context. Perhaps `globalLogger` might be more appropriate?

TidepoolKitUI/TidepoolKitUI.swift
```diff
@@ -0,0 +1,34 @@
...
+public class TidepoolKitUI {
+    
+    private var tpKit: TidepoolKit
+    
+    public init(tpKit: TidepoolKit, logger: TPKitLogging? = nil) {
+        clientLogger = logger
+        self.tpKit = tpKit
+    }
+    
+    public func logInViewController(loginSignupDelegate: LoginSignupDelegate? = nil, defaultServerHost: String?) -> UIViewController {
```
darinkrauss: [nit] Rename `defaultServerHost` -> `serverHost` since that is the name of the property on the `LoginViewController`.

TidepoolKitUI/TidepoolKitUI.swift
```diff
@@ -0,0 +1,34 @@
+//
+//  TidepoolKitUI.swift
+//  TidepoolKit
```
darinkrauss: [nit] TidepoolKitUI?

TidepoolKitUI/TidepoolKitUI.h
```diff
@@ -0,0 +1,19 @@
+//
+//  TidepoolKitUI.h
+//  TidepoolKit
```
darinkrauss: [nit] TidepoolKitUI?

TidepoolKitUI/LoginViewController.swift
```diff
@@ -0,0 +1,206 @@
+//
+//  LoginViewController.swift
+//  TidepoolKit
```
darinkrauss: [nit] TidepoolKitUI?

TidepoolKitUI/LoginViewController.swift
```diff
@@ -0,0 +1,206 @@
...
+//
+//  Created by Larry Kenyon on 8/23/19.
+//  Copyright © 2019 Tidepool Project. All rights reserved.
+//
+
+import UIKit
+import TidepoolKit
+
+
```
darinkrauss: [nit] Please remove all double blank lines with a single blank link throughout all files with a quick find/replace. I believe we decided that this extra spacing was not desired.

TidepoolKitUI/LoginViewController.swift
```diff
@@ -0,0 +1,206 @@
...
+public protocol LoginSignupDelegate: AnyObject {
+
+    func loginSignupComplete(_ session: TPSession)
+    func loginSignupCancelled()
+}
+
+
+class LoginViewController: UIViewController {
+
+    var loginSignupDelegate: LoginSignupDelegate?
```
darinkrauss: I think this should be `weak`?

TidepoolKitUI/LoginViewController.swift
```diff
@@ -0,0 +1,206 @@
...
+    var tpKit: TidepoolKit!
```
darinkrauss: [nit] `tidepoolKit`?

TidepoolKitUI/LoginViewController.swift
```diff
@@ -0,0 +1,206 @@
...
+                loginButton.setTitleColor(UIColor.black, for:UIControl.State())
+            }
+        } else {
+            loginButton.isEnabled = false
+            loginButton.setTitleColor(UIColor.gray, for:UIControl.State())
+        }
+    }
+    
+    private var currentServerHost: String {
+        return self.serverHost ?? tpKit.currentServerHost
```
darinkrauss: I don't think this should be reaching into `tidepoolKit` to get the current server host. The UI should be passed the server host to be used as the initial value. Then, this UI should maintain whatever is currently selected while the UI is visible. Finally, when the UI completes and calls the delegate it should pass back the final selected server host.

TidepoolKitUI/LoginViewController.swift
```diff
@@ -0,0 +1,206 @@
...
+    @IBOutlet weak var loginIndicator: UIActivityIndicatorView!
+    @IBOutlet weak var networkOfflineLabel: UILabel!
+    
+    override func viewDidLoad() {
+        super.viewDidLoad()
+        
+        configureForReachability()
+        updateButtonStates()
+        let notificationCenter = NotificationCenter.default
+        notificationCenter.addObserver(self, selector: #selector(LoginViewController.textFieldDidChange), name: UITextField.textDidChangeNotification, object: nil)
```
darinkrauss: Why are you using the notification center for the text field did change? Wouldn't it be more localized if this view controller were to implement the `UITextFieldDelegate` and handle it that way?

TidepoolKitUI/LoginViewController.swift
```diff
@@ -0,0 +1,206 @@
...
+        notificationCenter.addObserver(self, selector: #selector(LoginViewController.reachabilityChanged(_:)), name: ReachabilityChangedNotification, object: nil)
+        
+        self.configureCurrentServerButton()
```
darinkrauss: No `self.` needed here. Plus, I'd put the button state/configuration code together (eg. up before `updateButtonStates`).

TidepoolKitUI/LoginViewController.swift
```diff
@@ -0,0 +1,206 @@
...
+    }
+    
+    static var firstTime = true
+    override func viewDidAppear(_ animated: Bool) {
+        super.viewDidAppear(animated)
+        if LoginViewController.firstTime {
```
darinkrauss: Not sure what the purpose is here. Could you explain?

TidepoolKitExample/AppDelegate.swift
```diff
@@ -0,0 +1,46 @@
+//
+//  AppDelegate.swift
+//  TidepoolKit
```
darinkrauss: [nit] TidepoolKitExample?

TidepoolKit/Networking/APIConnect.swift
```diff
@@ -0,0 +1,950 @@
...
+    // TODO: if replacing a current NetworkRequestHandler, might want to cancel any outstanding requests!
+    func configureNetworkInterface(_ networkInterface: TidepoolNetworkInterface? = nil) {
+        if let networkInterface = networkInterface {
+            self.networkRequestHandler = networkInterface
+        } else {
+            self.networkRequestHandler = NetworkRequestHandler(apiQueue)
+        }
+    }
+    
+    private let kSessionTokenHeaderId = "X-Tidepool-Session-Token"
```
darinkrauss: The Loop project does not use the `k` prefix for constants. I suggest `tidepoolSessionTokenHeaderKey`. 

More importantly, we shouldn't have two different constants here either. We cannot assume the server will always respond with all lowercase (as HTTP header keys are case insensitive and the code needs to accommodate that). So, just define the canonical header (`private let tidepoolSessionTokenHeaderKey = "X-Tidepool-Session-Token") and then do lowercase manipulations, if necessary.

For example, the code below:
```
guard let token = httpResponse.allHeaderFields[self.kSessionTokenResponseId] as? String else
```
Could become something like:
```
guard let token = httpResponse.allHeaderFields.first(where: { return ($0 as? String)?.lowercased == tidepoolSessionTokenHeaderKey.lowercased })?.key as? String else
```

(I'm sure you can make that code cleaner, but you get the idea.)

TidepoolKit/Networking/APIConnect.swift
```diff
@@ -0,0 +1,950 @@
...
+            .data(using: String.Encoding.utf8.rawValue)?
+            .base64EncodedString(options: NSData.Base64EncodingOptions(rawValue: 0))
+        let headers = ["Authorization" : "Basic " + base64LoginString!]
+        
+        sendRequest("POST", urlExtension: urlExtension, contentType: .urlEncoded, headers:headers, requiresToken: false) {
+            sendResponse, error in
+            
+            guard error == nil else {
+                let error = error!
+                LogError("Login post failed with error: \(error)!")
```
darinkrauss: I don't think you need all of these `LogError` statements followed by a `completion(.failure...`). If the client received an error on completion then it can decide if it wants to log it or not. The other Loop libraries are not "chatty". It is up to the caller of the library whether to log anything or not.

TidepoolKit/Networking/APIConnect.swift
```diff
@@ -0,0 +1,950 @@
...
+                completion(Result.failure(error))
```
darinkrauss: [nit] You can remove the `Result` here. So, just `completion(.failure(error))`. Same throughout the rest of the code.

TidepoolKit/Networking/APIConnect.swift
```diff
@@ -0,0 +1,950 @@
...
+        if let networkInterface = networkInterface {
+            self.networkRequestHandler = networkInterface
+        } else {
+            self.networkRequestHandler = NetworkRequestHandler(apiQueue)
+        }
+    }
+    
+    private let kSessionTokenHeaderId = "X-Tidepool-Session-Token"
+    private let kSessionTokenResponseId = "x-tidepool-session-token"
+    let kDefaultServerHost = "stg-api.tidepool.org"
```
darinkrauss: This needs to be remove. As far as I'm concerned, there should be no default server host. It should be a required parameter by the client.

TidepoolKit/Networking/APIConnect.swift
```diff
@@ -0,0 +1,950 @@
...
+
+typealias NetworkRequestCompletionHandler = (Data?, URLResponse?, Error?) -> Void
+
+protocol TidepoolNetworkInterface {
+    func sendStandardRequest(_ request: URLRequest, completion: @escaping NetworkRequestCompletionHandler)
+    func sendBackgroundRequest(_ request: URLRequest, body: Data, completion: @escaping NetworkRequestCompletionHandler)
+}
+
+public protocol ReachabilitySource {
+    func serviceIsReachable() -> Bool
```
darinkrauss: Why not just call this `isReachable`. Then `Reachability` already implements it and you don't need another function that does the same thing.

TidepoolKit/Networking/APIConnect.swift
```diff
@@ -0,0 +1,950 @@
...
+}
+
+extension Reachability: ReachabilitySource {
+    public func serviceIsReachable() -> Bool {
+        return self.isReachable
+    }
+    public func configureNotifier(_ on: Bool) -> Bool {
+        if on {
+            do {
+                try self.startNotifier()
```
darinkrauss: In general, the Loop projects doe not use `self.` prefix except under 3 conditions: 1) in the initializer when *setting* instance properties; 2) when *setting* instance properties where `self.` is required for clarify in other places than an initializer; 3) when in a closure that requires `self.` (and it should be defined as `weak`).

TidepoolKit/Networking/APIConnect.swift
darinkrauss: Specifically, just:
```
var isReachable: Bool
```

TidepoolKit/Networking/APIConnect.swift
darinkrauss: Understood. Since it is not obvious (to me, at least), please replace occurrences of `LogInfo("")` with a reasonable message such as `LogInfo("Initializing APIConnector")` or something like that. It will help logging and well as code readability.

TidepoolKit/Networking/APIConnect.swift
```diff
@@ -0,0 +1,950 @@
...
+
+/// A singleton instance of APIConnector has the main responsibility of communicating to the Tidepool service:
+/// - Given a username and password, login.
+/// - Can refresh connection.
+/// - Provides online/offline status.
+/// - Get/put user data of various types (cbg, carb, etc.)
+class APIConnector {
+    
+    /// Current session: non-nil when "loggedIn", nil when "loggedOut"
+    var session: TPSession?
```
darinkrauss: I thought the session was going to be provided by the client when calling any API so TidepoolKit was more-or-less stateless?

TidepoolKit/Networking/APIConnect.swift
```diff
@@ -0,0 +1,950 @@
...
+            }
+        
+            self.session = TPSession(token, user: serviceUser, serverHost: serverHost)
+            LogInfo("Logged into \(serverHost) successfully! Returned userId = \(serviceUser.userId), userName: \(String(describing: serviceUser.userEmail))")
+            NotificationCenter.default.post(name: TidepoolLogInChangedNotification, object:self)
+            completion(Result.success(self.session!))
+        }
+    }
+    
+    func login(with session: TPSession) -> Result<TPSession, TidepoolKitError> {
```
darinkrauss: In general, why do some things have a `TP` prefix and others have `TidepoolKit` prefix?

TidepoolKit/Networking/APIConnect.swift
```diff
@@ -0,0 +1,950 @@
...
+        if self.session != nil {
```
darinkrauss: `guard session == nil else {` ?

darinkrauss (changes requested):
Some new comments so you can start responding and/or updating while I continue on with other comments.

TidepoolKit/Networking/APIConnect.swift
lakenyon: Done.

TidepoolKit/Networking/APIConnect.swift
lakenyon: Done.

TidepoolKitUITests/TidepoolKitUITests.swift
lakenyon: Done.

TidepoolKitUI/TidepoolKitUI.swift
lakenyon: Done.

TidepoolKitUI/TidepoolKitUI.swift
lakenyon: Changed this for TidepoolKit and TIdepoolKitUI files. Left it for test files, I find it more readable.

TidepoolKitUI/TidepoolKitUI.swift
lakenyon: True. I often do this so the reader doesn't need to check whether this is a local vs instance variable. Changed here.

TidepoolKitUI/TidepoolKitUI.swift
lakenyon: Changed. (Meaning was this was a logger coming from the framework client).

TidepoolKit/Networking/APIConnect.swift
lakenyon: I've learned to use it in more cases, but fine to elide it where not absolutely required (I do find it makes the code reading easier, especially when you aren't familiar with it).

lakenyon: Done.

TidepoolKit/Networking/APIConnect.swift
lakenyon: Generally TP prefix is what I've been using. I did use TidepoolKit as a prefix for the logging protocol and error type, changed both now to TP. Basically TP is the shortened version of TidepoolKit when the name gets long, but probably good to be consistent. So TidepoolKit will be used for the project names, and the two classes TidepoolKit and TidepoolKitUI; TP will be used for the public files associated with TidepoolKit.

TidepoolKit/Networking/APIConnect.swift
lakenyon: Yes, fixed.

TidepoolKitUI/TidepoolKitUI.swift
lakenyon: Done.

TidepoolKitUI/TidepoolKitUI.h
lakenyon: Hmmm... thought this was supposed to be the project name. Fixed.

TidepoolKitUI/LoginViewController.swift
lakenyon: Done.

TidepoolKitUI/LoginViewController.swift
lakenyon: Done.

TidepoolKitUI/LoginViewController.swift
lakenyon: Done.

TidepoolKitUI/LoginViewController.swift
lakenyon: Done.

TidepoolKitUI/LoginViewController.swift
lakenyon: It already passes back the selected server host in the session object. The problem is that there are two "default" server hosts: the last selected one (for debug login) and the production host, which is not returned by the dynamic DNS SRV record lookup. 

So I changed TidepoolKit to be ignorant of any of these, and to just require a server host to be passed into login. I also changed TidepoolKitUI to ensure that the production server host appears in the debug UI host selection list (test scenario where a non-production host has been selected previously, but now the user wants to go back to production, but the DNS lookup does not provide this). The client of TidepoolKitUI will now also have to conjure up a server host if it has not yet persisted one: this should be production for shipping code. 

I'm not overly fond of this solution because now the production server host needs to be hardwired in each client of TidepoolKitUI as well as in TidepoolKitUI itself (before this change it was just kept in TidepoolKit) - so, open to suggestions here...

TidepoolKitUI/LoginViewController.swift
lakenyon: Using notifications is an anachronism. Changed to just hook up the editing changed connection.

TidepoolKitUI/LoginViewController.swift
lakenyon: Done.

TidepoolKitUI/LoginViewController.swift
lakenyon: Removed, this was old code.

TidepoolKitExample/AppDelegate.swift
lakenyon: Done.

TidepoolKit/Networking/APIConnect.swift
lakenyon: Ok, changed to just use the lower-cased version, and lower-cased the checks on the response coming back. Removed all k prefixing from constants.

TidepoolKit/Networking/APIConnect.swift
lakenyon: Changed, see above.

TidepoolKit/Networking/APIConnect.swift
lakenyon: Done.

TidepoolKit/Networking/APIConnect.swift
lakenyon: Removed some of these that won't affect ability to debug issues.

TidepoolKit/Networking/APIConnect.swift
lakenyon: Done.

TidepoolKit/Networking/APIConnect.swift
lakenyon: Change in progress (fairly major, still adjusting all the test code)...

Btw, I thought we had just agreed that TidepoolKit would not persist any data itself. Even with this change there is state: the logger, any of the test protocol objects, any call in progress. The service itself has this state of a token being valid or not.

lakenyon: Ok, now most calls require a TPSession parameter, and APIConnect no longer has this session state. Calls relating to "log-in" state have been removed, along with the notification of log-in state change.

lakenyon:
Closing this initial PR, and creating a new one for everyone.

PR #2, opened 2019-10-16
Initial Version TidepoolKit, TidepoolKitUI frameworks.

lakenyon:


TidepoolKit/Third Party/Reachability.swift
```diff
@@ -0,0 +1,299 @@
+/*
```
ps2, line 1: The only real use for Reachability APIs is when deciding whether to retry a network connection, or to adjust error messaging.  Neither of those seem to be the uses here.  From what I can see, it's being used to prevent requests from being made, and prevent the user from engaging in login or other network activities. Both of those will prevent desired actions, when Reachability returns a false negative (network is powered down, but a request would trigger it to come up and would be successful).

ps2: I'd recommend we remove all the reachability code from this framework.

ps2 (changes requested):

darinkrauss:
Closing in favor of multiple smaller PRs.

PR #3, opened 2020-01-11
LOOP-951 Implement empty TidepoolKit project

darinkrauss:
@ps2 @nhamming This is pretty simple and doesn't really require both of you to approve. Just a basic starter project.


nhamming (approved):
I didn't detect any issues, but I do have one question. Should this be loaded as a plugin? If yes, should that be added now?

darinkrauss:
Hmm, good question, but I don't think so. Another project, `TidepoolServiceKit`, will implement the actual Loop service and Loop plugin. This project, `TidepoolKit` is just the common API/networking code that can be used by any iOS project. (In other words, `TidepoolKit` is to `TidepoolServiceKit` as `dexcom-icgm-sdk-ios` is to `dexcom-icgm-loop-plugin`.)

Does that make sense?

PR #4, opened 2020-02-20
LOOP-1/LOOP-9 Implement base networking and authentication & UI

darinkrauss:
- LOOP-1 Implement TidepoolKit base networking
- LOOP-9 Implement TidepoolKit authentication & UI
- TAPI interface to Tidepool API
- DNS SRV record lookup for non-production Tidepool environments
- Client configurable HTTP session
- Default User-Agent for all HTTP requests
- Basic networking functionality for all future functionality
- Placeholder login UI (ala Tidepool Mobile)
- Login, refresh, and logout functionality
- Enumerated errors
- Logging interface
- TidepoolKit Example app that utilizes TAPI


TidepoolKit Example/RootTableViewController.swift
```diff
@@ -0,0 +1,186 @@
...
+            return 1
+        case .authentication:
+            return Authentication.allCases.count
+        }
+    }
+
+    override func tableView(_ tableView: UITableView, cellForRowAt indexPath: IndexPath) -> UITableViewCell {
+        switch Section(rawValue: indexPath.section)! {
+        case .status:
+            let cell = tableView.dequeueReusableCell(withIdentifier: "StatusTableViewCell", for: indexPath) as! StatusTableViewCell
```
nhamming: [nit] Typically the cell is registered with the table view and reuse identifier uses `IdentifiableClass` `classname`.

TidepoolKit Example/RootTableViewController.swift
```diff
@@ -0,0 +1,186 @@
...
+                cell.environmentLabel?.text = session.environment.description
+                cell.environmentLabel?.lineBreakMode = .byTruncatingMiddle
+                cell.authenticationTokenLabel?.text = session.authenticationToken
+                cell.authenticationTokenLabel?.lineBreakMode = .byTruncatingMiddle
+                cell.userIDLabel?.text = session.userID
+            } else {
+                cell.stateLabel?.text = NSLocalizedString("Unauthenticated", comment: "The state text label an authenticated session does not exist")
+                cell.environmentLabel?.text = NSLocalizedString("-", comment: "The environment label placeholder when an authenticated session does not exist")
+                cell.authenticationTokenLabel?.text = NSLocalizedString("-", comment: "The authentication token label placeholder when an authenticated session does not exist")
+                cell.userIDLabel?.text = NSLocalizedString("-", comment: "The user id label placeholder when an authenticated session does not exist")
```
nhamming: [nit] this no value string could be defined statically with the cell.

TidepoolKit Example/RootTableViewController.swift
```diff
@@ -0,0 +1,186 @@
...
+            }
+            return cell
+        case .authentication:
+            switch Authentication(rawValue: indexPath.row)! {
+            case .login:
+                let cell = tableView.dequeueReusableCell(withIdentifier: "LabelTableViewCell", for: indexPath) as! LabelTableViewCell
```
nhamming: [nit] not sure how visually consistent this example needs to be to loop, but I think there already exists a cell that acts like a button. Look at `TextButtonTableViewCell`.

nhamming (commented):
Some UI nits, otherwise I didn't detect anything.

TidepoolKit Example/RootTableViewController.swift
darinkrauss: TidepoolKit exists outside of the Loop ecosystem. It is intended to eventually be a common framework that easily allows any Swift code in any client to utilize the Tidepool API. In this case, this is also a simple example app that has really no relation to Loop. It primarily serves as my test bed for the TAPI functionality in TidepoolKit and secondarily as an example app for future consumers of the framework.

TidepoolKit Example/RootTableViewController.swift
darinkrauss: @nhamming True, but wouldn't I still need to use this text in the `prepareForReuse` function to reset it back to the default?

TidepoolKit Example/RootTableViewController.swift
darinkrauss: The `IdentifiableClass` protocol is provided by `LoopKit` and this project does not have access to `LoopKit` (by design since this is not a Loop-specific project). I think this is fine for this example class.

darinkrauss:
@nhamming My comments notwithstanding, I took a closer look at `TextButtonTableViewCell` from `LoopKit` and decided much of that made sense here. I copied over the relevant bits and updated appropriately. Would you please re-review the latest commit?

darinkrauss:
I added some further updates and refactoring to better support base networking in the next ticket.

TidepoolKit/Internal/DNS.swift
```diff
@@ -17,11 +17,32 @@ enum DNSError: Error {
...
+        if lhs.weight > rhs.weight {
+            return true
+        } else if lhs.weight < rhs.weight {
+            return false
+        }
+        if lhs.host < rhs.host {
+            return true
+        } else if lhs.host > rhs.host {
+            return false
+        }
```
nhamming, line 40: [nit] This portion of the comparison is confusing to me as I don't understand why the `host` construction is part of this comparison. The previous comparison where equal hosts use the port and `localhost` is being detected was easier for me to understand.

nhamming (dismissed):
I had confusion on the updated comparison, but otherwise LGTM.

TidepoolKit Example/RootTableViewController.swift
nhamming: Yeah, I take your point that this is just moving code. I was also suggesting that a single definition of `"-"` would make updates easier, but this is very minor. Truly a `nit` :)

darinkrauss: I updated to make the `"-"` definition common. Also moved the middle truncation settings into the storyboard.

TidepoolKit/Internal/DNS.swift
darinkrauss: This comparison is only used to order the environments in a rationale and stable manner. It uses priority and weight from the SRV record first. Then, if those are identical for two environments, it alphabetically compares the hosts. Then, if the hosts match, compares ports. 

I moved the check for `localhost` outside of this comparison since it is overriding the natural order to "fix" something the backend should change (in my opinion). I recommended to the backend that they update the DNS SRV records to actually order via priority and weight so *all* clients get the same rational ordering (production, integration, qa1, and qa2) and `localhost` would only be added manually by the clients if desired (not in the DNS SRV record since that should only be used to advertise environments hosted by the backed; `localhost` is hosted by the local development environment and shouldn't be seen by normal users even if they know the special way to switch environments).

darinkrauss: So, with that said, going to leave this as-is.

darinkrauss:
Since @ps2 has been occupied, adding @rickpasetto to review, particularly since TidepoolKit is not actually Loop-specific.

@rickpasetto Some background... TidepoolKit is intended to eventually be a high-level wrapper around the Tidepool backend API for iOS apps (though there will likely also be an Android one at some point). At this time, its focus is strictly on Loop, so only iPhone, iOS 13+, and limited to what Loop needs functionality-wise. This first version basically just includes authentication related items. Do note that login, refresh, logout, and UI functionality will be replaced with updated versions when the backend finishes their move to AWS Cognito. So, no unit tests for this since all of the authentication related code is temporary. The next PR after this one will have code for stable APIs so will include unit tests. Please ping me with any other questions. Thanks!

rickpasetto:
@darinkrauss cool beans.  I'm taking a look at it now.

TidepoolKit/TAPI.swift
```diff
@@ -0,0 +1,244 @@
...
+                            completion(.success(TSession(environment: environment, authenticationToken: authenticationToken, userID: loginResponse.userID)))
+                        } else {
+                            completion(.failure(.responseNotAuthenticated(response, data)))
+                        }
+                    } else {
+                        completion(.failure(.responseMalformedJSON(response, data)))
+                    }
+                } else {
+                    completion(.failure(.responseMissingJSON(response)))
+                }
```
rickpasetto, line 106: [nit] To avoid the Pyramid of Doom, consider changing this to `guard` statemetents.  e.g.:
```swift
                guard let data = data else {
                    completion(.failure(.responseMissingJSON(response)))
                    return
                }
                
                guard let dictionaryArray = try? JSONSerialization.jsonObject(with: data) as? [String: Any], let loginResponse = LoginResponse(rawValue: dictionaryArray) else {
                    completion(.failure(.responseMalformedJSON(response, data)))
                    return
                }
                
                guard let authenticationToken = response.value(forHTTPHeaderField: "X-Tidepool-Session-Token"), !authenticationToken.isEmpty else {
                    completion(.failure(.responseNotAuthenticated(response, data)))
                    return
                }
                
                completion(.success(TSession(environment: environment, authenticationToken: authenticationToken, userID: loginResponse.userID)))
```

TidepoolKit/TAPI.swift
```diff
@@ -0,0 +1,244 @@
...
+    ///
+    /// - Parameters:
+    ///   - environment: The environment to login.
+    ///   - email: The email to use for login.
+    ///   - password: The password to use for login.
+    ///   - completion: The completion function to invoke with the newly created session or an error.
+    public func login(environment: TEnvironment, email: String, password: String, completion: @escaping (Result<TSession, TError>) -> Void) {
+        var request = createRequest(environment: environment, method: "POST", path: "/auth/login")
+        request.setValue(basicAuthorizationFromCredentials(email: email, password: password), forHTTPHeaderField: "Authorization")
+        performRequest(request) { (result: Result<(HTTPURLResponse, Data?), TError>) -> Void in
```
rickpasetto: [nit] to help improve readability, consider making `Result<(HTTPURLResponse, Data?), TError>` a `typealias`?
(e.g. at the top of the file:
```swift
    typealias HTTPResult = Result<(HTTPURLResponse, Data?), TError>
```
Up to you if you want to scope it in this class or not...seems like it might be generally useful so I'd put it outside.

TidepoolKit/TAPI.swift
```diff
@@ -0,0 +1,244 @@
...
+            }
+        }
+    }
+
+    private func performRequest(_ request: URLRequest, completion: @escaping (Result<(HTTPURLResponse, Data?), TError>) -> Void) {
+        let task = urlSession.dataTask(with: request) { (data, response, error) in
+            if let error = error {
+                completion(.failure(.network(error)))
+            } else {
+                completion(self.processStatusCode(response: response as! HTTPURLResponse, data: data))
```
rickpasetto: Consider checking the type here instead of force-casting.
e.g.:
```swift
        guard let error = error else {
            completion(.failure(.network(error)))
            return
         }
         guard let resp = response as? HTTPURLResponse else {
            completion(.failure(.unexpectedResponse)) // <-- you'll need to define this yourself
            return
         }
         completion(self.processStatusCode(response: resp, data: data))
```

TidepoolKit/TAPI.swift
```diff
@@ -0,0 +1,244 @@
...
+        return "\(Bundle.main.userAgent) \(Bundle(for: self).userAgent) \(ProcessInfo.processInfo.userAgent)"
+    }
+
+    private static var defaultURLSessionConfiguration: URLSessionConfiguration {
+        let urlSessionConfiguration = URLSessionConfiguration.ephemeral
+        urlSessionConfiguration.waitsForConnectivity = true
+        if urlSessionConfiguration.httpAdditionalHeaders == nil {
+            urlSessionConfiguration.httpAdditionalHeaders = [:]
+        }
+        urlSessionConfiguration.httpAdditionalHeaders!["User-Agent"] = TAPI.defaultUserAgent
```
rickpasetto: [nit] This is minor, but my eyes always have to stare hard at any force-unwraps or force-casts.  Perhaps instead of force-unwrapping here, since you are pretty much guaranteeing `urlSessionConfiguration.httpAdditionalHeaders` is not `nil` in the lines above, maybe consider this:
```swift
        urlSessionConfiguration.httpAdditionalHeaders?["User-Agent"] = TAPI.defaultUserAgent
```

TidepoolKit/TAPI.swift
```diff
@@ -0,0 +1,244 @@
...
+    ///   - completion: The completion function to invoke with the latest environments or any error.
+    public func fetchEnvironments(completion: ((Result<[TEnvironment], TError>) -> Void)? = nil) {
+        DispatchQueue.global(qos: .background).async {
+            DNS.lookupSRVRecords(for: TAPI.DNSSRVRecordsDomainName) { result in
+                switch result {
+                case .failure(let error):
+                    if let completion = completion {
+                        completion(.failure(.network(error)))
+                    } else {
+                        TSharedLogging.error("Failure during DNS SRV record lookup: \(error)")
```
rickpasetto, line 53: Hm... just curious: why are you only logging if `completion` is `nil`?  Wouldn't you want to log regardless?
Then this code could just be:
```swift
                        completion?(.failure(.network(error)))
                        TSharedLogging.error("Failure during DNS SRV record lookup: \(error)")
```
?

TidepoolKit/TAPI.swift
```diff
@@ -0,0 +1,244 @@
...
+                            return record
+                        }
+                        return DNSSRVRecord(priority: UInt16.max, weight: record.weight, host: record.host, port: record.port)
+                    }
+                    let environments = records.sorted().environments
+                    self.environmentsLocked.mutate { $0 = environments }
+                    if let completion = completion {
+                        completion(.success(environments))
+                    } else {
+                        TSharedLogging.debug("Successful DNS SRV record lookup")
```
rickpasetto, line 68: Similar to line 53 (https://github.com/tidepool-org/TidepoolKit/pull/4/files#diff-7c0ec7bfb91406887343e411a4c96013R53) above...

rickpasetto (dismissed):
@darinkrauss Just a few minor comments, but other than that looks pretty good to me!

TidepoolKit/TAPI.swift
darinkrauss: This function is refactored in the next version when there is more than one type that needs to be deserialized from JSON. This pyramid ends up going away with that refactoring. Will leave this as-is considering that.

TidepoolKit/TAPI.swift
darinkrauss: :+1: I'll add a check for that. It shouldn't ever happen, but reasonable to check and not die.

TidepoolKit/TAPI.swift
darinkrauss: Fair enough. Changed.

TidepoolKit/TAPI.swift
darinkrauss: The intention is to leave it up to the completion handler to "do the right thing", whatever that is for the client.

TidepoolKit/TAPI.swift
darinkrauss: Same as above.

darinkrauss:
@nhamming or @rickpasetto I believe I've updated or responded to all of your comments. Can I get an approval? Thanks!

TidepoolKit/TAPI.swift
rickpasetto: Ok, so maybe I misunderstand what the purpose of `TSharedLogging` is.  Is it important to diagnose problems, regardless what the client does?

rickpasetto (approved):
I'm totally fine with this going in as @darinkrauss has updated it.  My additional comments are just questions that could be addressed later.

PR #5, opened 2020-03-06
LOOP-6/LOOP-4 Implement TidepoolKit upload of initial data types

darinkrauss:
- https://tidepool.atlassian.net/browse/LOOP-4
- https://tidepool.atlassian.net/browse/LOOP-6
- Get profile, list/create data sets, list/create data APIs
- Use Tidepool API compliant JSON encoding and decoding
- Defacto encoding and decoding for generic dictionary from JSON
- Implement base Tidepool API datum
- Initial CGM and Food datum types
- Add association, dictionary, location, and origin common structs
- Add blood glucose constants and target structs
- Add profile and contained patient structs
- Add dataset struct
- Updates to example app to reflect TAPI changes
- Add sharing to example app for debugging purposes
- Add sample data
- Refactor errors to capture all specific conditions
- Refactor request create and perform request to utilize Codable
- Replace all usage RawRepresentation with Codable
- Loads of tests

### DIFF
--- a/TidepoolKit/Internal/Extensions/Codable.swift
+++ b/TidepoolKit/Internal/Extensions/Codable.swift
@@ -1,0 +1,222 @@
+//
+//  Codable.swift
+//  TidepoolKit
+//
+//  Created by Darin Krauss on 2/26/20.
+//
+
+// Original: https://gist.github.com/sukov/d3834c0e7b72e4f7575f753b352f6ddd
+
+import Foundation
+
+struct JSONCodingKeys: CodingKey {
+    var stringValue: String
+
+    init(stringValue: String) {
+        self.stringValue = stringValue
+    }
+
+    var intValue: Int?
+
+    init?(intValue: Int) {
+        self.init(stringValue: "\(intValue)")
+        self.intValue = intValue
+    }
+}
+
+extension KeyedDecodingContainer {
+    func decode(_ type: [[String: Any]].Type, forKey key: K) throws -> [[String: Any]] {
+        var container = try nestedUnkeyedContainer(forKey: key)
+        return try container.decode(type)
+    }
+
+    func decode(_ type: [String: Any].Type, forKey key: K) throws -> [String: Any] {
+        let container = try nestedContainer(keyedBy: JSONCodingKeys.self, forKey: key)
+        return try container.decode(type)
+    }
+
+    func decode(_ type: [Any].Type, forKey key: K) throws -> [Any] {
+        var container = try nestedUnkeyedContainer(forKey: key)
+        return try container.decode(type)
+    }
+
+    func decodeIfPresent(_ type: [[String: Any]].Type, forKey key: K) throws -> [[String: Any]]? {
+        guard contains(key) else { return nil }
+
+        return try decode(type, forKey: key)
+    }
+
+    func decodeIfPresent(_ type: [String: Any].Type, forKey key: K) throws -> [String: Any]? {
+        guard contains(key) else { return nil }
+
+        return try decode(type, forKey: key)
+    }
+
+    func decodeIfPresent(_ type: [Any].Type, forKey key: K) throws -> [Any]? {
+        guard contains(key) else { return nil }
+
+        return try decode(type, forKey: key)
+    }
+
+    func decode(_ type: [String: Any].Type) throws -> [String: Any] {
+        var dictionary = [String: Any]()
+
+        for key in allKeys {
+            if let boolValue = try? decode(Bool.self, forKey: key) {
+                dictionary[key.stringValue] = boolValue
+            } else if let intValue = try? decode(Int.self, forKey: key) {
+                dictionary[key.stringValue] = intValue
+            } else if let stringValue = try? decode(String.self, forKey: key) {
+                dictionary[key.stringValue] = stringValue
+            } else if let doubleValue = try? decode(Double.self, forKey: key) {
+                dictionary[key.stringValue] = doubleValue
+            } else if let nestedDictionary = try? decode([String: Any].self, forKey: key) {
+                dictionary[key.stringValue] = nestedDictionary
+            } else if let nestedArray = try? decode([Any].self, forKey: key) {
+                dictionary[key.stringValue] = nestedArray
+            } else if let isValueNil = try? decodeNil(forKey: key), isValueNil == true {
+                dictionary[key.stringValue] = nil
+            } else {
+                throw DecodingError.dataCorrupted(DecodingError.Context(codingPath: codingPath + [key], debugDescription: "Unable to decode value"))
+            }
+        }
+        return dictionary
+    }
+}
+
+extension UnkeyedDecodingContainer {
+    mutating func decode(_ type: [[String: Any]].Type) throws -> [[String: Any]] {
+        var array: [[String: Any]] = []
+
+        while isAtEnd == false {
+            if let dictionary = try? decode([String: Any].self) {
+                array.append(dictionary)
+            } else {
+                throw DecodingError.dataCorrupted(DecodingError.Context(codingPath: codingPath, debugDescription: "Unable to decode value"))
+            }
+        }
+        return array
+    }
+
+    mutating func decode(_ type: [Any].Type) throws -> [Any] {
+        var array: [Any] = []
+
+        while isAtEnd == false {
+            if let value = try? decode(Bool.self) {
+                array.append(value)
+            } else if let value = try? decode(Int.self) {
+                array.append(value)
+            } else if let value = try? decode(String.self) {
+                array.append(value)
+            } else if let value = try? decode(Double.self) {
+                array.append(value)
+            } else if let nestedDictionary = try? decode([String: Any].self) {
+                array.append(nestedDictionary)
+            } else if let nestedArray = try? decodeNestedArray([Any].self) {
+                array.append(nestedArray)
+            } else if let isValueNil = try? decodeNil(), isValueNil == true {
+                array.append(Optional<Any>.none as Any)
+            } else {
+                throw DecodingError.dataCorrupted(DecodingError.Context(codingPath: codingPath, debugDescription: "Unable to decode value"))
+            }
+        }
+        return array
+    }
+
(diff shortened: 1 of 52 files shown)


TidepoolKit/Internal/Extensions/Codable.swift
```diff
@@ -0,0 +1,222 @@
...
+
+    var intValue: Int?
+
+    init?(intValue: Int) {
+        self.init(stringValue: "\(intValue)")
+        self.intValue = intValue
+    }
+}
+
+extension KeyedDecodingContainer {
```
rickpasetto, line 27: Maybe I just cannot find them, but where are the unit tests for these extensions?

TidepoolKit/Internal/Extensions/Date.swift
```diff
@@ -0,0 +1,70 @@
...
+
+    var timeString: String {
+        return Date.timeFormatter.string(from: self.roundedToTimeInterval(TimeInterval.millisecond))
+    }
+
+    var deviceTimeString: String {
+        return Date.deviceTimeFormatter.string(from: self.roundedToTimeInterval(TimeInterval.millisecond))
+    }
+
+    static let timeFormatter: ISO8601DateFormatter = {
```
rickpasetto, line 40: [nit] This is fine, but I was just wondering, are these alternate time formats associated with specific versions of the Tidepool API?  If so, it might be handy to either name these as such or maybe just comment.

TidepoolKit/Internal/Extensions/Date.swift
```diff
@@ -0,0 +1,70 @@
...
+//  Date.swift
+//  TidepoolKit
+//
+//  Created by Darin Krauss on 11/5/19.
+//  Copyright © 2019 Tidepool Project. All rights reserved.
+//
+
+import Foundation
+
+public extension Date {
```
rickpasetto: Unit tests for these?

TidepoolKit/Internal/Responses/DataResponse.swift
```diff
@@ -0,0 +1,55 @@
...
+            do {
+                var datum: TDatum?
+                let type = try superContainer.decode(TDatum.DatumType.self, forKey: .type)
+                switch type {
+                case .cbg:
+                    datum = try TCBGDatum(from: superDecoder)
+                case .food:
+                    datum = try TFoodDatum(from: superDecoder)
+                default:
+                    break // DEPRECATED: Ignore
```
rickpasetto, line 33: Seems if this is not an error, we should probably at least log.  If, someday, the backend adds a type, and someone has an old version of the app installed, there'd be no indication of this.  My recommendation, though, is to throw an error in this case.

TidepoolKit/Internal/Responses/DataResponse.swift
```diff
@@ -0,0 +1,55 @@
...
+struct DataResponse: Codable {
+    var data: [TDatum]
+    var malformed: [Int: [String: Any]]
+
+    init(data: [TDatum] = [], malformed: [Int: [String: Any]] = [:]) {
+        self.data = data
+        self.malformed = malformed
+    }
+    
+    init(from decoder: Decoder) throws {
```
rickpasetto, line 18: Unit tests?

TidepoolKit/Internal/Responses/DataResponse.swift
```diff
@@ -0,0 +1,55 @@
+//
+//  DataResponse.swift
+//  TidepoolKit
+//
+//  Created by Darin Krauss on 2/28/20.
+//  Copyright © 2020 Tidepool Project. All rights reserved.
+//
+
+struct DataResponse: Codable {
+    var data: [TDatum]
```
rickpasetto, line 10: General comment about all of the `Codable` model objects: I notice you use `var`s for their members...is it your intent that they should be mutable like this?  I'm curious why they are not all `let`s.

TidepoolKit/TAPI.swift
```diff
@@ -84,25 +83,25 @@ public class TAPI {
...
-                            completion(.failure(.responseNotAuthenticated(response, data)))
-                        }
+                switch error {
+                case .requestNotAuthorized(let response, let data):     // CUSTOM: Backend currently returns status code 403 to imply email not verified
+                    completion(.failure(.requestEmailNotVerified(response, data)))
+                default:
+                    completion(.failure(error))
+                }
+            case .success(let response, let data, let loginResponse):
+                if let authenticationToken = response.value(forHTTPHeaderField: "X-Tidepool-Session-Token"), !authenticationToken.isEmpty {
```
rickpasetto, line 97: [nit] consider putting this header key into a constant

TidepoolKit/TAPI.swift
```diff
@@ -145,40 +144,199 @@ public class TAPI {
...
+        return request
+    }
+
+    private func createRequest(session: TSession, method: String, path: String, queryItems: [URLQueryItem]? = nil) -> URLRequest? {
         var request = createRequest(environment: session.environment, method: method, path: path, queryItems: queryItems)
-        request.setValue(session.authenticationToken, forHTTPHeaderField: "X-Tidepool-Session-Token")
+        request?.setValue(session.authenticationToken, forHTTPHeaderField: "X-Tidepool-Session-Token")
         if let trace = session.trace {
-            request.setValue(trace, forHTTPHeaderField: "X-Tidepool-Trace-Session")
+            request?.setValue(trace, forHTTPHeaderField: "X-Tidepool-Trace-Session")
```
rickpasetto: Similar comment regarding a constant for header field key

TidepoolKit/TAPI.swift
```diff
@@ -145,40 +144,199 @@ public class TAPI {
...
+                completion(nil)
+            }
+        }
+    }
+
+    // MARK: - Internal - Create Request
+
+    private func createRequest<E>(session: TSession, method: String, path: String, body: E) -> URLRequest? where E: Encodable {
+        var request = createRequest(session: session, method: method, path: path)
+        request?.setValue("application/json; charset=utf-8", forHTTPHeaderField: "Content-Type")
```
rickpasetto: consider putting `contentType` as a constant

rickpasetto (commented):
🌟 for committing to `Codable` !  Good job.

Some general thoughts and comments:
1. Great job on the unit tests! I also would love to see some tests on the extensions and some other smaller classes you created (comments were added inline)
2. I'm not sure why Model objects need `var` members.  Prefer immutable objects where you can.
3. You've had to implement a lot of HTTP request/response code on your own here.  Did you consider using a library such as [AlamoFire](https://github.com/Alamofire/Alamofire) to help?  (it would save you from having to write unit tests for all if it too).

Oh one last thing: consider enabling unit test code coverage in the test scheme.

nhamming (approved):
Lots of code, but nothing jumped out at me.

TidepoolKit/Internal/Responses/DataResponse.swift
darinkrauss: Good question. 

One thing I struggled with was dealing with all (17) of the properties common to all Tidepool data types (as implemented in the TDatum base class). If I implemented properties as `let` then every single property would have to be specified in every initializer for all Tidepool data types. This got very unreadable and annoying very fast. 

I decided the ability to create the object using a straightforward, type-specific constructor (with only the few type-specific properties) and then allowing modification of the objects afterwards (from `var`) to set any of the infrequently-used common properties made the most sense.

In addition, while writing code to use these classes, I realized that there is the added benefit of being able to set these common properties en-mass after creating a number of data.

While it is not ideal I think this is a reasonable compromise.

TidepoolKit/Internal/Responses/DataResponse.swift
darinkrauss: Sorry, the comment is misleading. The default case is only because I did not include all of the implemented Tidepool data model types in this PR. Once all types are included (in subsequent PRs) there will be no default case and any unexpected type (aka a new or renamed type) will throw an error when decoding the `enum`. The thrown error will be captured and the "bad" data will be added to the malformed list (unparsable for whatever reason) that can be logged by the client.

TidepoolKit/Internal/Extensions/Date.swift
darinkrauss: These two Date formats cover all of the Tidepool data model and are not version dependent. However, I just noticed the extension is `public` and need not be. I will change to avoid cluttering the public namespace.

TidepoolKit/Internal/Extensions/Codable.swift
darinkrauss: Frankly, I'm not a fan of testing internal implementation details, within which this is included. This is akin to testing `private` functions and properties on a struct/class.

In general, my feeling is that if the public-facing "API" is fully and properly tested then the internal implementation should also be tested.

However, if you feel strongly about unit tests here I will implement them. @rickpasetto

TidepoolKit/Internal/Extensions/Date.swift
darinkrauss: I removed the accidental inclusion of the `public` attribute on this extension. I did not provide unit tests per not testing internal implementation details, as above.

TidepoolKit/Internal/Responses/DataResponse.swift
darinkrauss: I did not provide unit tests per not testing internal implementation details, as above.

rickpasetto (approved):
LGTM 👍

TidepoolKit/Internal/Responses/DataResponse.swift
rickpasetto: I admit that adding all those initializers is a drag.  In the past, I've done exactly what you've done: started with `var`s because it was such a pain.  In the end, I eventually came around and realized that immutability is actually more beneficial to the users, and it was just boilerplate code anyway.  That said, not a blocker for this review, but something to chew on...

TidepoolKit/Internal/Extensions/Codable.swift
rickpasetto: I do feel strongly about it, @darinkrauss ,  but it shouldn't block your PR.  We should discuss it in another forum.

PR #6, opened 2020-03-13
Various updates to prepare for additional data types and bug fixes

darinkrauss:
- https://tidepool.atlassian.net/browse/LOOP-4
- https://tidepool.atlassian.net/browse/LOOP-6
- Replace unnecessary Meta struct with generic TDictionary
- Check for nil response from DNS query
- Refactor common carbohydrate and insulin units
- Refactor common blood glucose start target
- Add data set id to status area in example app
- Remove redundant status state in example app
- Fix login UI images in dark mode
- Fix bug with malformed data dictionary encoding

@rickpasetto Let me know if this is a good size for a PR. Happy to change if not. Thanks!


TidepoolKit/Model/TBloodGlucose.swift
```diff
@@ -37,4 +37,35 @@ public struct TBloodGlucose {
             self.high = high
         }
     }
+
+    public struct StartTarget: Codable, Equatable {
+        public var start: Int?
+        public var target: Double?
+        public var range: Double?
```
nhamming, line 44: hm... I find this confusing. The range is a `Double`, but I don't see how a range is made from this value. Should a comment be included?

TidepoolKit/Model/TInsulin.swift
```diff
@@ -0,0 +1,13 @@
...
+//  TidepoolKit
+//
+//  Created by Darin Krauss on 3/9/20.
+//  Copyright © 2020 Tidepool Project. All rights reserved.
+//
+
+public struct TInsulin {
+    public enum Units: String, Codable {
+        case units = "Units"
+    }
```
nhamming, line 12: hm... I would have expected `IU` (international units). Does Units = IU? Are these synonymous in this space?

nhamming (approved):
Just a couple of questions that I don't intend to be blockers (since it may be that I just have a misunderstanding).

TidepoolKit/Internal/DNS.swift
```diff
@@ -48,7 +48,9 @@ class DNS {
 
         let serviceRef: UnsafeMutablePointer<DNSServiceRef?> = UnsafeMutablePointer.allocate(capacity: MemoryLayout<DNSServiceRef>.size)
         let callback: DNSServiceQueryRecordReply = { (_, _, _, _, _, _, _, rdlen, rdata, _, context) -> Void in
-            context?.assumingMemoryBound(to: DNSSRVRecordHandler.self).pointee(DNSSRVRecord(data: Data(bytes: rdata!, count: Int(rdlen))))
+            if let rdata = rdata, rdlen > 0 {
+                context?.assumingMemoryBound(to: DNSSRVRecordHandler.self).pointee(DNSSRVRecord(data: Data(bytes: rdata, count: Int(rdlen))))
+            }
```
rickpasetto, line 64: [nit] else... log an error or something? Admittedly I'm not sure what to do in that case, just wondering if you considered the `else` case here.

TidepoolKit/Model/TBloodGlucose.swift
```diff
@@ -37,4 +37,35 @@ public struct TBloodGlucose {
             self.high = high
         }
     }
+
+    public struct StartTarget: Codable, Equatable {
+        public var start: Int?
```
rickpasetto, line 42: I'm also a bit confused by this.  If `start` is non-optional in the constructors, why is it optional here?  Is it valid to be nil? (not blocking the review, just asking your consideration of it)

TidepoolKit/Internal/Responses/DataResponse.swift
```diff
@@ -8,9 +8,9 @@
 
 struct DataResponse: Codable {
     var data: [TDatum]
-    var malformed: [Int: [String: Any]]
+    var malformed: [String: [String: Any]]
 
-    init(data: [TDatum] = [], malformed: [Int: [String: Any]] = [:]) {
+    init(data: [TDatum] = [], malformed: [String: [String: Any]] = [:]) {
```
rickpasetto: [nit] Is this `[String: [String: Any]]` structure a common "malformed" type?  Perhaps consider a typealias for it?

rickpasetto (approved):
Added some non-blocking or "nit" comments, but other than that LGTM.

TidepoolKit/Model/TBloodGlucose.swift
darinkrauss: These fields are directly from the Tidepool data model and are documented here: https://developer.tidepool.org/data-model/device-data/types/pumpSettings.html#bgtarget. There are certain combinations of these fields that make sense for certain devices. Specifically, range can be used with either target or high and they mean different things on the different devices. (The joys of trying to standardize diabetes data from many different manufacturers.)

In any case, rather than try to duplicate any documentation from the Tidepool data model, I've chosen to have no documentation here (as it will just grow stale over time) and simply expect that anyone interested will also be looking at the Tidepool data model documentation. Since all of this code "mirrors" the backend API and, thus, the Tidepool data model, I think that is the safest bet. In short, I don't expect anyone to simply take up this code and start uploading data. They will have to understand how this data is used first via the Tidepool data model documentation.

TidepoolKit/Model/TInsulin.swift
darinkrauss: As I understand, the term "Unit" (with this capitalization) is the officially recognized unit of insulin measurement and what should be used for all scholarly papers, technical writing, etc. I can find the source if you'd like, but it's what we've been using for a while now.

TidepoolKit/Model/TBloodGlucose.swift
darinkrauss: This is another one of those tradeoffs. These data types will be used to both write data (Loop, others) and read data (Loop Follow, others) from the Tidepool backend. 

If a client is using this library to write data, then it will end up using the initializers, so it requires the `start` property. It is absolutely required by current backend validations and will fail if it is missing.

However, this will also be used to read any and all data from the Tidepool backend. Unfortunately, not all data returned by the Tidepool backend is entirely clean. That is, I know for certain that there is data in the database that does not conform to all of the current validations. (Yes, there is a tech-debt ticket to go through and "cleanup/validate" all existing data, but it is still outstanding and is not a high priority compared to getting Loop running.)

So, to be lenient with this non-conformant data, you'll notice that almost every property is optional (except for the barest minimum to allow the API to function; specifically type, subType/deliveryType, and time).

While not ideal, it is the best possible solution in my mind without: 1) trying to duplicate all validations from the backend, and 2) failing every single slightest validation error on legacy data from the backend.

rickpasetto: I buy that.

darinkrauss:
@rickpasetto Would you please review just the latest commit (af7eb115f9bb13bc04f3ae172e90cbe83480eca4) for the updates I made based upon your PR review? Thanks!

rickpasetto (approved):
Looks good to me!  🚢

PR #7, opened 2020-03-14
Add applicationSettings, bloodKetone, reportedState, SMBG, and water types

darinkrauss:
- Add datum selector for delete operations


nhamming (approved):
LGTM

rickpasetto (approved):
Nice!  💯 Looks good to me!

I'm not an expert on the API, so I can't speak to whether this is a correct binding, but the code looks good.

PR #8, opened 2020-03-18
Add delete data API; update food datum with estimated absorption duration

darinkrauss:
- Add delete data API
- Move separate TDatumSelector to TDatum.Selector
- Add estimatedAbsorptionDuration to food datum
- Add application origin type
- Use unique origin id in all sample data
- Capture datum selector from previously created data in example


TidepoolKit/TAPI.swift
```diff
@@ -250,6 +255,23 @@ public class TAPI {
...
+    ///   - dataSetId: The data set id from which to delete the data.
+    ///   - session: The Tidepool API session to use.
+    ///   - completion: The completion function to invoke with any error.
+    public func deleteData(withSelectors selectors: [TDatum.Selector], dataSetId: String, session: TSession, completion: @escaping (TError?) -> Void) {
+        guard !selectors.isEmpty else {
+            completion(nil)
+            return
+        }
+        
+        let request = createRequest(session: session, method: "DELETE", path: "/v1/data_sets/\(dataSetId)/data", body: selectors)
```
rickpasetto, line 271: [nit] unit test for this?

rickpasetto (approved):
Just one request for a unit test, but other than that LGTM.

rickpasetto:
👍 Nice, @darinkrauss .  Not sure if you are looking for another PR, but LGTM! 🚢

TidepoolKit/Model/Datum/TFoodDatum.swift
```diff
@@ -79,12 +79,14 @@ public class TFoodDatum: TDatum, Decodable {
         public var fat: Fat?
         public var protein: Protein?
         public var energy: Energy?
+        public var estimatedAbsorptionDuration: Int?
```
ps2, line 82: Probably better as a TimeInterval.  Or if not, can units be added into the variable name?

ps2 (approved):
LGTM, just a suggestion on making estimated duration a TimeInterval object.

TidepoolKit/Model/Datum/TFoodDatum.swift
darinkrauss: I actually had a trial implementation using `TimeInterval`. I agree, it does make more sense. However, I realized that I'd have to change 11 different structs that currently use the default `Codable` implementation with a custom implementation just to support `TimeInterval`. 

The backend, for better or worse, requires the value to be an integer (while `TimeInterval` is a `Double` at heart) and, depending upon the struct, the value might be in units of minutes, seconds, or milliseconds. (Yea, the API design needs a serious consistency overhaul.)

I thought about changing the field name to add a `Minutes`, `Seconds`, or `Milliseconds` prefix, but, there again, it also requires updating those same 11 different structs with custom code.

Finally, I thought about adding documentation, but there are lots of places that would benefit from documentation. I really, really want to stay out of the business of keeping the TidepoolKit documentation up-to-date with the Tidepool API documentation (which is the source of truth). In my mind it is almost a requirement to have the Tidepool API documentation up when using TidepoolKit. There are too many caveats and situations that are impossible to codify and simply require written documentation.

So, at this point I'm going to pass on changing this PR. However, I'd like to keep the conversation going and if we decide to change something, I can do a final pass over all data types once all of the PRs are through.

PR #9, opened 2020-03-22
Add cgmSettings and pumpSettings types

darinkrauss:
- https://tidepool.atlassian.net/browse/LOOP-883


TidepoolKit/Model/Datum/TCGMSettingsDatum.swift
```diff
@@ -0,0 +1,240 @@
...
+        try container.encodeIfPresent(scheduledAlerts, forKey: .scheduledAlerts)
+        try container.encodeIfPresent(highAlertsDEPRECATED, forKey: .highAlertsDEPRECATED)
+        try container.encodeIfPresent(lowAlertsDEPRECATED, forKey: .lowAlertsDEPRECATED)
+        try container.encodeIfPresent(outOfRangeAlertsDEPRECATED, forKey: .outOfRangeAlertsDEPRECATED)
+        try container.encodeIfPresent(rateOfChangeAlertsDEPRECATED, forKey: .rateOfChangeAlertsDEPRECATED)
+        try super.encode(to: encoder)
+    }
+
+    public struct Alerts: Codable, Equatable {
+        public var enabled: Bool?
```
nhamming, line 71: Curious what this enabled refers to? Does this mean that the alerts are enabled? If so, at least with the Dexcom integration, the mapping of alerts to enabled is 1-to-1 and some alerts cannot be disabled. Or does this refer to always sound is enabled? Then the mapping of alerts to enabled is many-to-1, but again some alerts always sounds cannot be disabled (if critical alerts has been allowed by the user). The Tidepool Loop model has both an enabled and always sound enabled property for each alert.

TidepoolKit Example/Support/Extensions/Sample.swift
```diff
@@ -21,6 +21,23 @@ struct Sample {
...
+                                  manufacturers: ["Acme Pump", "Acme CGM"],
+                                  model: "Super Deluxe",
+                                  serialNumber: "999666333",
+                                  transmitterId: "A1B2C3",
+                                  units: .milligramsPerDeciliter,
+                                  defaultAlerts: cgmSettingsAlerts,
+                                  scheduledAlerts: [TCGMSettingsDatum.ScheduledAlert(name: "Default",
+                                                                                     days: ["sunday", "monday", "tuesday", "wednesday", "thursday", "friday", "saturday"],
+                                                                                     start: 0,
+                                                                                     end: 64800000,
```
nhamming, line 34: Is this is seconds or milliseconds? Either way is `64800000` a good value?

TidepoolKit Example/Support/Extensions/Sample.swift
```diff
@@ -64,6 +110,21 @@ struct Sample {
...
+                                            urgentLow: TCGMSettingsDatum.Alerts.LevelAlert(enabled: true, level: 55, units: .milligramsPerDeciliter, snooze: snooze),
+                                            urgentLowPredicted: TCGMSettingsDatum.Alerts.LevelAlert(enabled: true, level: 55, units: .milligramsPerDeciliter, snooze: snooze),
+                                            low: TCGMSettingsDatum.Alerts.LevelAlert(enabled: true, level: 70, units: .milligramsPerDeciliter, snooze: snooze),
+                                            lowPredicted: TCGMSettingsDatum.Alerts.LevelAlert(enabled: true, level: 65, units: .milligramsPerDeciliter, snooze: snooze),
+                                            high: TCGMSettingsDatum.Alerts.LevelAlert(enabled: true, level: 180, units: .milligramsPerDeciliter, snooze: snooze),
+                                            highPredicted: TCGMSettingsDatum.Alerts.LevelAlert(enabled: true, level: 200, units: .milligramsPerDeciliter, snooze: snooze),
+                                            fall: TCGMSettingsDatum.Alerts.RateAlert(enabled: true, rate: 3, units: .milligramsPerDeciliterPerMinute, snooze: snooze),
```
nhamming, line 122: This seems like a strange value for a `fall`. Is `+3` really the expected slope of the curve?

TidepoolKit Example/Support/Extensions/Sample.swift
```diff
@@ -64,6 +110,21 @@ struct Sample {
...
+                                            rise: TCGMSettingsDatum.Alerts.RateAlert(enabled: true, rate: 3, units: .milligramsPerDeciliterPerMinute, snooze: snooze),
+                                            noData: TCGMSettingsDatum.Alerts.DurationAlert(enabled: true, duration: 30, units: .minutes, snooze: snooze),
```
nhamming, line 124: hm... curious about this alert as well. `noData` seems to implies that the communication is broken. For the Dexcom implementation we have `noReadings` which implies that communication exists but the sensor cannot provide a reading for the current state. Does `noData` map to `noReadings`? If yes, consider renaming.

TidepoolKit Example/Support/Extensions/Sample.swift
```diff
@@ -64,6 +110,21 @@ struct Sample {
                                         energy: TFoodDatum.Nutrition.Energy(350, .kilocalories))
         }
 
+        fileprivate static var cgmSettingsAlerts: TCGMSettingsDatum.Alerts {
+            let snooze = TCGMSettingsDatum.Alerts.Snooze(30, .minutes)
+            return TCGMSettingsDatum.Alerts(enabled: true,
```
nhamming, line 115: Is this intended to only represent the configurable alerts?  I ask since the only configurable part of the `urgentLow` alert (maybe the `urgentLowPredicted` alert too, but I haven't implemented this in Tidepool Loop) is the sound. I don't see any sounds included here. So makes me wonder if `urgentLow` should be included. If yes, then should non-configured alerts also be included? It is totally possible various CGMs will have various non-configurable alerts. Do we need to track these alerts such that we can recreate a specific scenario?

TidepoolKit/Model/Datum/TCGMSettingsDatum.swift
```diff
@@ -0,0 +1,240 @@
...
+
+import Foundation
+
+public class TCGMSettingsDatum: TDatum, Decodable {
+    public typealias Units = TBloodGlucose.Units
+
+    public var manufacturers: [String]?
+    public var model: String?
+    public var serialNumber: String?
+    public var transmitterId: String?
```
nhamming, line 17: Not sure how granular the settings need to be, but do we want to include sensor information in these settings (use of a code or no code)?

How about the transmitter configuration (software version, firmware version, backfill supported, max run time, etc.)?

TidepoolKit/Model/Datum/TCGMSettingsDatum.swift
```diff
@@ -0,0 +1,240 @@
...
+                self.level = level
+                self.units = units
+                self.snooze = snooze
+            }
+        }
+
+        public struct RateAlert: Codable, Equatable {
+            public enum Units: String, Codable {
+                case milligramsPerDeciliterPerMinute = "mg/dL/minute"
+                case millimolesPerLiterPerMinute = "mmol/L/minute"
```
nhamming, line 136: Could this not just use the units of the CGM settings and add `/minute`?

TidepoolKit/Model/Datum/TCGMSettingsDatum.swift
```diff
@@ -0,0 +1,240 @@
...
+            public init(enabled: Bool, duration: Double, units: Units, snooze: Snooze? = nil) {
+                self.enabled = enabled
+                self.duration = duration
+                self.units = units
+                self.snooze = snooze
+            }
+        }
+
+        public struct LevelAlert: Codable, Equatable {
+            public typealias Units = TBloodGlucose.Units
```
nhamming, line 118: Should these units be allowed to be different than the CGM settings units?

TidepoolKit/Model/Datum/TCGMSettingsDatum.swift
```diff
@@ -0,0 +1,240 @@
...
+            }
+        }
+    }
+
+    public struct ScheduledAlert: Codable, Equatable {
+        public var name: String?
+        public var days: [String]?
+        public var start: Int?
+        public var end: Int?
+        public var alerts: Alerts?
```
nhamming, line 174: At least for the Dexcom implementation, the alert schedule can be enabled or disabled (as well as always sound enabled/disabled).

nhamming (changes requested):
Took a close look at the CGM settings and reviewed the Pump settings. Possible I have some misunderstanding, but if not changes are needed to better represent the Dexcom implementation of alerts.

rickpasetto (commented):
I'm not sure if this is just a first-pass at trying to match existing Tidepool backend schema, or if this is trying to match Loop's CGM settings parameters.  If it is the latter, @nhamming has some good questions, and I think I'll defer to his expertise here.

TidepoolKit Example/Support/Extensions/Sample.swift
darinkrauss: Per backend documentation, this is milliseconds and can be up to 24 hours. The maximum `end` value is `86400000` (24 hours * 60 minutes/hour * 60 seconds/minute * 1000 milliseconds/second), so, yes, this is a valid value and represents 6pm.

TidepoolKit Example/Support/Extensions/Sample.swift
darinkrauss: The `TCGMSettingsDatum` represents the *current* Tidepool Data Model and API and was mostly modeled after the Dexcom API. Specifically, this endpoint: https://developer.dexcom.com/get-devices

I have not yet made any changes that might be needed to represent Tidepool Loop Dexcom CGM settings. We should start a conversation on how to get those CGM settings into `Loop`/`LoopKit` proper and then to `TidepoolService` and `TidepoolKit` so we can figure out what, if any, changes we need to make to the Tidepool Data Model and API. (As far as I am aware, the Dexcom G6 CGM settings are still "private" to the plugin.)

In response to your specific questions, this represents what the Dexcom API sends to Tidepool backend. If I understand correctly, it does represent that latest configuration (not defaults) and includes user-settable (e.g. high alert) and fixed alerts (e.g. urgent low).

In our future conversation, we should discuss user-settable and fixed alerts as well as sounds.

Another thing we should consider is whether the settings of each plugin are part of the official documented Tidepool Data Model and API or whether we have a "private black box" data blob for saving and restoring this configuration.

TidepoolKit Example/Support/Extensions/Sample.swift
darinkrauss: The `3` represents the setting the user chooses in the Dexcom UI which is either `2` or `3` mg/dL/min. It is the absolute value of actual slope.

TidepoolKit Example/Support/Extensions/Sample.swift
darinkrauss: The term `noData` comes from the Dexcom API itself. Specifically, according to Dexcom documentation:

> "noData" – triggers when the display device stops receiving EGV data from the transmitter for a certain amount of time; corresponds to the No Readings alert

TidepoolKit/Model/Datum/TCGMSettingsDatum.swift
darinkrauss: We can definitely expand upon these when we update the Tidepool Data Model and API later.

TidepoolKit/Model/Datum/TCGMSettingsDatum.swift
darinkrauss: The backend is doing all validation enforcement. Much like documenting, I really don't want to get into the business of having to keep the validations in sync between this code and the backend.

I suppose I could remove these units, but this is reflecting an actually used and deployed API, not anything brand new. So, going to leave this as-is.

TidepoolKit/Model/Datum/TCGMSettingsDatum.swift
darinkrauss: In theory, yes, but that would require a complete `Codable` implementation for `RateAlert` to support that here. Not sure that is worth the effort.

TidepoolKit/Model/Datum/TCGMSettingsDatum.swift
darinkrauss: It was promoted from the `ScheduledAlert` struct below.

TidepoolKit/Model/Datum/TCGMSettingsDatum.swift
darinkrauss: It was promoted to the `Alerts ` struct above.

darinkrauss:
@nhamming @rickpasetto Would you both please re-review given the comments in this PR and the context of this work as I described in the #loop-dev channel. I don't see anything here that isn't specifically about the new Tidepool Loop Dexcom G6 integration (which I would strongly prefer we discuss and figure out as a separate issue that I can implement *after* all of these PRs).

TidepoolKit/Model/Datum/TCGMSettingsDatum.swift
```diff
@@ -0,0 +1,240 @@
...
+                self.duration = duration
+                self.units = units
+            }
+        }
+    }
+
+    public struct ScheduledAlert: Codable, Equatable {
+        public var name: String?
+        public var days: [String]?
+        public var start: Int?
```
rickpasetto, line 172: [nit] `start` and `end` are... `Int`s?  What do they represent?  Seconds since midnight?  Maybe a comment to that effect?

rickpasetto (approved):
@darinkrauss you answered my question in https://tidepoolteam.slack.com/archives/CG1MZQRJ5/p1585085586489700.  Sounds like this is a first pass at representing Tidepool's `cgmSettings` API, and that another pass may be needed to normalize these with what Loop needs.  So, understanding that, this looks ok to me.  (I've added a few comments, not-blocking though)

TidepoolKit Example/Support/Extensions/Sample.swift
nhamming: I suspect Dexcom G6 app converts this value from what is displayed in the UI based on my knowledge of the SDK. Tidepool Loop needs to provide a negative value when constructing the actual alert. Is basing this on the UI correct? Wouldn't it be better to provide the actual value instead of its absolute value? Or is the absolute value always provided for such things to the backend? I see potential for confusion.

TidepoolKit Example/Support/Extensions/Sample.swift
nhamming: That's my point. The alert is the `noReadings` alert. Again I see potential for confusion here by calling it `noData`.

nhamming: Is there a reason to keep it `noData`?

TidepoolKit Example/Support/Extensions/Sample.swift
nhamming: Is there a reason that the absolute value is needed here?

nhamming (approved):
I don't really understand Tidepool backend or Dexcom web API. Otherwise, looks good to me.

ps2 (approved):
LGTM!

TidepoolKit Example/Support/Extensions/Sample.swift
darinkrauss: Well, this is already a published, public API and there is already data in the backend database using this format. So, I don't think we can (easily) change this going now without requiring a backend change and a database migration. Plus, potentially updating any client apps that already use this data.

In short, while not idea, I think this has to stay as-is. I don't think it is worth the effort to change at this point.

TidepoolKit Example/Support/Extensions/Sample.swift
darinkrauss: Again, a published and public API, so changing is unlikely. Besides, there are a whole huge number of "naming differences" between Loop and the Tidepool data model. This is just something to understand when the TidepoolService, the "translation layer" from Loop to Tidepooo data model, is reviewed.

TidepoolKit/Model/Datum/TCGMSettingsDatum.swift
darinkrauss: As I've mentioned elsewhere, this user of this code will need to keep the Tidepool data model documentation on the side when implementation code with TidepoolKit. There are so, so many caveats, validations, and limits that cannot easily be expressed in this code and I really don't want to try to keep any sort of documentation here up-to-date with the "truth" backend documentation.

PR #10, opened 2020-03-25
Add insulin and physicalActivity types

darinkrauss:

### DIFF
--- a/TidepoolKit Example/Support/Extensions/Sample.swift
+++ b/TidepoolKit Example/Support/Extensions/Sample.swift
@@ -47,6 +47,26 @@ struct Sample {
                            mealOther: "midnight snack",
                            nutrition: foodNutrition,
                            ingredients: [foodIngredient, foodIngredient]),
+                TInsulinDatum(time: Date(),
+                              dose: TInsulinDatum.Dose(active: 1.23,
+                                                       correction: 2.34,
+                                                       food: 3.45,
+                                                       total: 4.56),
+                              formulation: insulinFormulation,
+                              site: "Forehead"),
+                TPhysicalActivityDatum(time: Date(),
+                                       name: "Exercise",
+                                       aggregate: true,
+                                       activityType: .other,
+                                       activityTypeOther: "extremeSports",
+                                       duration: TPhysicalActivityDatum.Duration(1.2, .hours),
+                                       distance: TPhysicalActivityDatum.Distance(5.0, .kilometers),
+                                       elevationChange: TPhysicalActivityDatum.ElevationChange(500, .meters),
+                                       flight: TPhysicalActivityDatum.Flight(count: 160),
+                                       lap: TPhysicalActivityDatum.Lap(1, TPhysicalActivityDatum.Distance(5.0, .kilometers)),
+                                       step: TPhysicalActivityDatum.Step(count: 12345),
+                                       reportedIntensity: .high,
+                                       energy: TPhysicalActivityDatum.Energy(800, .kilocalories)),
                 TPumpSettingsDatum(time: Date(),
                                    activeScheduleName: "Normal",
                                    basal: TPumpSettingsDatum.Basal(rateMaximum: TPumpSettingsDatum.Basal.RateMaximum(3),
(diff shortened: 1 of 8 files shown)


TidepoolKit Example/Support/Extensions/Sample.swift
```diff
@@ -47,6 +47,26 @@ struct Sample {
                            mealOther: "midnight snack",
                            nutrition: foodNutrition,
                            ingredients: [foodIngredient, foodIngredient]),
+                TInsulinDatum(time: Date(),
+                              dose: TInsulinDatum.Dose(active: 1.23,
+                                                       correction: 2.34,
+                                                       food: 3.45,
+                                                       total: 4.56),
```
nhamming, line 54: [nit] I get that these are just arbitrary numbers, but should `total` be the sum of those above? Or is `total` something else?

TidepoolKit/Model/Datum/TInsulinDatum.swift
```diff
@@ -0,0 +1,117 @@
...
+//  Created by Darin Krauss on 11/1/19.
+//  Copyright © 2019 Tidepool Project. All rights reserved.
+//
+
+import Foundation
+
+public class TInsulinDatum: TDatum, Decodable {
+    public var dose: Dose?
+    public var formulation: Formulation?
+    public var site: String?
```
nhamming, line 14: Can you really have just a formulation or site without a dose?

TidepoolKit/Model/Datum/TPhysicalActivityDatum.swift
```diff
@@ -0,0 +1,265 @@
...
+        case preparationAndRecovery
+        case racquetball
+        case rowing
+        case rugby
+        case running
+        case sailing
+        case skatingSports
+        case snowboarding
+        case snowSports
+        case soccer
```
nhamming, line 66: [nit] Interesting that there is `americanFootball` and `autralianFootball`, but only `soccer` and not `football`.

TidepoolKit/Model/Datum/TPhysicalActivityDatum.swift
```diff
@@ -0,0 +1,265 @@
...
+    }
+
+    public var name: String?
+    public var aggregate: Bool?
+    public var activityType: ActivityType?
+    public var activityTypeOther: String?
+    public var duration: Duration?
+    public var distance: Distance?
+    public var elevationChange: ElevationChange?
+    public var flight: Flight?
```
nhamming, line 103: [nit] I personally find `flight` confusing and don't really know what I would put for a value.

nhamming (approved):
One comment and rest nits. Otherwise LGTM

ps2 (approved):
LGTM, in terms of style and structure. I haven't reviewed for conformance to the Tidepool API.

TidepoolKit Example/Support/Extensions/Sample.swift
darinkrauss: Total is how much was delivered. You can have some currently `active`, some *recommended* for `correction` to address a BG above target, some *recommended* for the `food` you just ate, and still dose a `total` different than the calculated `correction + food - active`.

TidepoolKit/Model/Datum/TInsulinDatum.swift
darinkrauss: No, you should not generally *create* a `TInsulinDatum` without a `dose`, which is why the initializer requires the `dose`. However, when parsing pre-existing data from the server, it is possible, though highly unlikely, to not have a dose.

TidepoolKit/Model/Datum/TPhysicalActivityDatum.swift
darinkrauss: When the backed data model for physical activity was created, I used the activities as found in HealthKit.

TidepoolKit/Model/Datum/TPhysicalActivityDatum.swift
darinkrauss: Again, this came from HealthKit, though I see it is called "totalFlightsClimbed".  This `flight` properties is a `Flight` struct which just contains a `count` int.

darinkrauss: Perhaps not ideal, but it is what is published and public.

PR #11, opened 2020-03-26
Add dosingDecision and pumpStatus types

darinkrauss:
- https://tidepool.atlassian.net/browse/LOOP-883


TidepoolKit/Model/Datum/TDosingDecisionDatum.swift
```diff
@@ -0,0 +1,145 @@
...
+//  TDosingDecisionDatum.swift
+//  TidepoolKit
+//
+//  Created by Darin Krauss on 3/6/20.
+//  Copyright © 2020 Tidepool Project. All rights reserved.
+//
+
+import Foundation
+
+public class TDosingDecisionDatum: TDatum, Decodable {
```
rickpasetto, line 11: I just noticed...why are these `class`es and not `struct`s?  If you stick with structs, copy semantics are clearer (and optimized in Swift).  Mixing structs and classes can lead to confusion.  I suggest making everything structs if possible.

rickpasetto (commented):
Question about class vs struct.  Other than that this LGTM.

TidepoolKit/Model/Datum/TDosingDecisionDatum.swift
darinkrauss: Anything with a `*Datum` prefix is a subclass of `TDatum` and in some cases you have an intermediate subclass (e.g. `TScheduledBasalDatum` is a subclass of `TBasalDatum` which is a subclass of `TDatum`). 

This goes back to our discussion some time ago about how to represent all of the primary Tidepool data model types. We discussed enums (with up to three structs for each case), embedded structs, and a class hierarchy. After much back-and-forth (and writing experimental code for each choice) I ended up with the class hierarchy because it most closely modeled the actual data model.

However, you'll notice that all supporting structures are `structs` for the very reason you mention.

I know, it is not ideal, but I felt it was the best of the various options.

rickpasetto: Sorry, right.

rickpasetto (approved):
LGTM!

nhamming (approved):
LGTM

PR #13, opened 2020-04-03
Add all device event types

darinkrauss:

### DIFF
--- a/TidepoolKit/Model/Datum/DeviceEventDatum/TAlarmDeviceEventDatum.swift
+++ b/TidepoolKit/Model/Datum/DeviceEventDatum/TAlarmDeviceEventDatum.swift
@@ -1,0 +1,56 @@
+//
+//  TAlarmDeviceEventDatum.swift
+//  TidepoolKit
+//
+//  Created by Darin Krauss on 11/1/19.
+//  Copyright © 2019 Tidepool Project. All rights reserved.
+//
+
+import Foundation
+
+public class TAlarmDeviceEventDatum: TDeviceEventDatum, Decodable {
+    public enum AlarmType: String, Codable {
+        case autoOff = "auto_off"
+        case lowInsulin = "low_insulin"
+        case lowPower = "low_power"
+        case noDelivery = "no_delivery"
+        case noInsulin = "no_insulin"
+        case noPower = "no_power"
+        case occlusion
+        case other
+        case overLimit = "over_limit"
+    }
+
+    public var alarmType: AlarmType?
+    public var status: TStatusDeviceEventDatum?     // NOTE: API write only
+    public var statusId: String?                    // NOTE: API read only
+
+    public init(time: Date, alarmType: AlarmType, status: TStatusDeviceEventDatum? = nil) {
+        self.alarmType = alarmType
+        self.status = status
+        super.init(.alarm, time: time)
+    }
+
+    public required init(from decoder: Decoder) throws {
+        let container = try decoder.container(keyedBy: CodingKeys.self)
+        self.alarmType = try container.decodeIfPresent(AlarmType.self, forKey: .alarmType)
+        self.statusId = try container.decodeIfPresent(String.self, forKey: .status)
+        try super.init(.alarm, from: decoder)
+    }
+
+    public override func encode(to encoder: Encoder) throws {
+        var container = encoder.container(keyedBy: CodingKeys.self)
+        try container.encodeIfPresent(alarmType, forKey: .alarmType)
+        if let status = status {
+            try container.encodeIfPresent(status, forKey: .status)
+        } else if let statusId = statusId {
+            try container.encodeIfPresent(statusId, forKey: .status)
+        }
+        try super.encode(to: encoder)
+    }
+
+    private enum CodingKeys: String, CodingKey {
+        case alarmType
+        case status
+    }
+}
--- a/TidepoolKit/Model/Datum/DeviceEventDatum/TReservoirChangeDeviceEventDatum.swift
+++ b/TidepoolKit/Model/Datum/DeviceEventDatum/TReservoirChangeDeviceEventDatum.swift
@@ -1,0 +1,39 @@
+//
+//  TReservoirChangeDeviceEventDatum.swift
+//  TidepoolKit
+//
+//  Created by Darin Krauss on 11/1/19.
+//  Copyright © 2019 Tidepool Project. All rights reserved.
+//
+
+import Foundation
+
+public class TReservoirChangeDeviceEventDatum: TDeviceEventDatum, Decodable {
+    public var status: TStatusDeviceEventDatum?     // NOTE: API write only
+    public var statusId: String?                    // NOTE: API read only
+
+    public init(time: Date, status: TStatusDeviceEventDatum? = nil) {
+        self.status = status
+        super.init(.reservoirChange, time: time)
+    }
+
+    public required init(from decoder: Decoder) throws {
+        let container = try decoder.container(keyedBy: CodingKeys.self)
+        self.statusId = try container.decodeIfPresent(String.self, forKey: .status)
+        try super.init(.reservoirChange, from: decoder)
+    }
+
+    public override func encode(to encoder: Encoder) throws {
+        var container = encoder.container(keyedBy: CodingKeys.self)
+        if let status = status {
+            try container.encodeIfPresent(status, forKey: .status)
+        } else if let statusId = statusId {
+            try container.encodeIfPresent(statusId, forKey: .status)
+        }
+        try super.encode(to: encoder)
+    }
+
+    private enum CodingKeys: String, CodingKey {
+        case status
+    }
+}
--- a/TidepoolKit/Model/Datum/DeviceEventDatum/TTimeChangeDeviceEventDatum.swift
+++ b/TidepoolKit/Model/Datum/DeviceEventDatum/TTimeChangeDeviceEventDatum.swift
@@ -1,0 +1,81 @@
+//
+//  TTimeChangeDeviceEventDatum.swift
+//  TidepoolKit
+//
+//  Created by Darin Krauss on 11/1/19.
+//  Copyright © 2019 Tidepool Project. All rights reserved.
+//
+
+import Foundation
+
+public class TTimeChangeDeviceEventDatum: TDeviceEventDatum, Decodable {
+    public enum Method: String, Codable {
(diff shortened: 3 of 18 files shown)


TidepoolKit/Model/Datum/DeviceEventDatum/TTimeChangeDeviceEventDatum.swift
```diff
@@ -0,0 +1,81 @@
...
+        try super.encode(to: encoder)
+    }
+
+    public struct Info: Codable, Equatable {
+        public var time: String?
+        public var timeZoneName: String?
+
+        public init(time: String? = nil, timeZoneName: String? = nil) {
+            self.time = time
+            self.timeZoneName = timeZoneName
```
ps2, line 53: I'm not aware of any pumps that actually have timezone capability; they run on a 24 hour clock that is only changed when the user or controller changes the time. The time on events reported from the pump can be converted to a universal time by using a fixed offset.  Using an actual timezone for this conversion will lead to errors, when the timezone changes its offset, but the pump doesn't.  Is the plan here to store fixed offset timezone names, like "UTC−06:00"? If so, we should document which format is expected here, since there are two unfortunately counterintuitive conventions: "ETC/GMT+6" refers to the same fixed offset timezone as "UTC−06:00". Notice the sign is different.

ps2 (approved):
LGTM! I just have a question about a `timeZoneName` field.

TidepoolKit/Model/Datum/DeviceEventDatum/TTimeChangeDeviceEventDatum.swift
darinkrauss: From the Tidepool data model perspective "devices" aren't limited to just pumps. Most of the device event sub types are specific to pumps, but they don't have to be. In this case, we are using the `deviceEvent/timeChange` to capture the time zone change for the mobile device itself for Tidepool Mobile. This is what we will be using to capture time zone changes for the iOS device for Tidepool Loop as we discussed in yesterday's meeting (as opposed to adding the time zone on each data point).

ps2: How is a pump time change recorded?

darinkrauss: For all of the existing pumps via Tidepool Uploader, it uses the same `deviceEvent/timeChange` datum, but does not use the time zone field, but rather the time field. So, if someone changes their Medtronic pump from 11:12:34am to 12:10:00pm on 4/3/2020, then the event would look something like:
```
{
  "type": "deviceEvent",
  "subType": "timeChange",
  "from": {
    "time": "2020-04-03T11:12:34"
  },
  "to": {
    "time": "2020-04-03T12:10:00"
  }
}
```
Furthermore, these events are really "informational" only. The Uploader takes these known pump time changes and updates all of the UTC time fields in all data retrieved from a pump to create a coherent UTC time ordered stream of data that is uploaded to the backend. This process is called "UTC Bootstrapping" and is documented here: https://github.com/tidepool-org/uploader/blob/master/docs/BootstrappingToUTC.md

The only "device" that uses the `deviceEvent/timeChange` datum with time zones is a device running Tidepool Mobile. Tidepool Loop will use this as well to indicate that the user's device has changed time zones.

darinkrauss: Oh, and the time zones that Tidepool Mobile report (and Tidepool Loop will report) are the actual names (not offset based names) such as "America/Los_Angeles".

ps2: Thanks for explaining how this field is used. That all makes sense, I think.

In addition to device time change, and user timezone change, there is a third item that is probably worth discussing a bit here:

Loop tracks a utc offset that is used for inferring schedule inflection points, and inferring pump event times.  This offset can change when the user syncs time in Loop, if the current user's timezone offset is different than the pump's utc offset. This changes independently of the user's timezone changing. All time inferences using Loop's schedules need to use this pump utc offset. I would think we *do* have some need to make inferences using Loop's configured schedules on the backend, so we need a way to track them. I assume those will be uploaded with schedules (which should trigger a change record when the utc offset is changed).

If that assumption is correct, then I don't have any more questions.  Thanks again, Darin.

darinkrauss: If I understand correctly, I think I've already accounted the third item. 😄 

In the Tidepool data model `pumpSettings` type (which is populated from `StoredSettings` in Tidepool Loop), there is a new-for-Tidepool-Loop field called `scheduleTimeZoneOffset` that accounts for the pump UTC offset. 

(It used to be split into 4 fields in a previous PR - one for each schedule: basal rates, bg targets, carb ratios, and ISFs - but I have since collapsed that into one common field for all schedules and that collapsed changed will be in a future PR).

Any time the Loop settings are changed a new `StoredSettings` would be saved off and a new `pumpSettings` (among other things) would be uploaded to the backend.

Does that make sense and does that cover your concern?

ps2: Yes, sounds like what you describe matches my assumption. Thanks!

TidepoolKit/Model/Datum/DeviceEventDatum/TAlarmDeviceEventDatum.swift
```diff
@@ -0,0 +1,56 @@
...
+//  TidepoolKit
+//
+//  Created by Darin Krauss on 11/1/19.
+//  Copyright © 2019 Tidepool Project. All rights reserved.
+//
+
+import Foundation
+
+public class TAlarmDeviceEventDatum: TDeviceEventDatum, Decodable {
+    public enum AlarmType: String, Codable {
```
rickpasetto, line 12: I'm unfamiliar with the Tidepool API so please forgive my ignorance.  This type, `TAlarmDeviceEventDatum`, seems to be generically-named to apply to any "device" alarms, but the `AlarmType`s seem to me to only apply to insulin devices (indeed, this seems to be true of the other "DeviceEvent" types... more applicable to insulin devices than CGMs).  Why is that?  Again, I'm just asking here.

TidepoolKit/Model/Datum/DeviceEventDatum/TReservoirChangeDeviceEventDatum.swift
```diff
@@ -0,0 +1,39 @@
...
+//  TidepoolKit
+//
+//  Created by Darin Krauss on 11/1/19.
+//  Copyright © 2019 Tidepool Project. All rights reserved.
+//
+
+import Foundation
+
+public class TReservoirChangeDeviceEventDatum: TDeviceEventDatum, Decodable {
+    public var status: TStatusDeviceEventDatum?     // NOTE: API write only
```
rickpasetto, line 12: Question: What do these notes mean?  That users of `TidepoolKit` should treat these as write [read] only?

TidepoolKit/Model/Datum/DeviceEventDatum/TReservoirChangeDeviceEventDatum.swift
```diff
@@ -0,0 +1,39 @@
...
+    public var statusId: String?                    // NOTE: API read only
```
rickpasetto, line 13: Question: Is this a generic id, or should this be a UUID?

TidepoolKit/Model/Datum/DeviceEventDatum/TTimeChangeDeviceEventDatum.swift
rickpasetto: I just want to add that I +1 the notion of putting some documentation comment in here that `timeZoneName` expects an "official" time zone name, please.  IANA does define them (https://www.iana.org/time-zones).  Even special cases like Arizona (which does not observe DST changes) and the Navajo nation (which does, despite being in Arizona) are accounted for there.  (also see https://en.wikipedia.org/wiki/List_of_tz_database_time_zones)

rickpasetto (commented):
@darinkrauss Overall looks good to me, but first I have some (hopefully) quick questions...

TidepoolKit/Model/Datum/DeviceEventDatum/TAlarmDeviceEventDatum.swift
darinkrauss: Yes, *currently* `TAlarmDeviceEventDatum` is particularly focused on insulin delivery devices. That's primarily just because historically we haven't had other devices report specific alarms. However, with Tidepool Loop actively managing CGMs I can see we will probably want to expand the Tidepool data model to accommodate these new alarms/errors. We have yet to figure out what will need to be reported so these changes will come some time after this PR.

TidepoolKit/Model/Datum/DeviceEventDatum/TReservoirChangeDeviceEventDatum.swift
darinkrauss: Yes, but as I've indicated before the implementor needs to have the existing Tidepool data model documentation handy for more information.

TidepoolKit/Model/Datum/DeviceEventDatum/TReservoirChangeDeviceEventDatum.swift
darinkrauss: This is a Tidepool datum id. Again, it is all documented in the Tidepool data model docs.

TidepoolKit/Model/Datum/DeviceEventDatum/TReservoirChangeDeviceEventDatum.swift
rickpasetto: ok thanks

TidepoolKit/Model/Datum/DeviceEventDatum/TReservoirChangeDeviceEventDatum.swift
rickpasetto: ok thanks

rickpasetto (approved):
LGTM!

PR #14, opened 2020-04-06
Add all bolus and calculator types

darinkrauss:


rickpasetto (approved):
LGTM
(Note I'm mostly only checking for syntax, not for API accuracy...)

TidepoolKit/Model/Datum/BolusDatum/TBolusDatum.swift
```diff
@@ -0,0 +1,50 @@
...
+        case extended = "square"
+        case normal
+    }
+
+    public let subType: SubType
+    public var insulinFormulation: InsulinFormulation?
+
+    public init(_ subType: SubType, time: Date, insulinFormulation: InsulinFormulation? = nil) {
+        self.subType = subType
+        self.insulinFormulation = insulinFormulation
```
nhamming, line 25: [nit] will the `insulinFormulation` ever be set? Seems like the subclasses don't include this information?

TidepoolKit/Model/Datum/TCalculatorDatum.swift
```diff
@@ -0,0 +1,100 @@
...
+    public var insulinOnBoard: Double?
+    public var bloodGlucoseInput: Double?
+    public var insulinSensitivity: Double?
+    public var carbohydrateInput: Double?
+    public var insulinCarbohydrateRatio: Double?
+    public var bloodGlucoseTarget: BloodGlucoseTarget?
+    public var recommended: Recommended?
+    public var bolus: TBolusDatum?      // NOTE: API write only
+    public var bolusId: String?         // NOTE: API read only
+    public var units: Units?
```
nhamming, line 24: [nit] is units too generic, since there are a number of different measurements/observations included?

TidepoolKit/Model/Datum/TCalculatorDatum.swift
```diff
@@ -0,0 +1,100 @@
...
+
+    public var insulinOnBoard: Double?
+    public var bloodGlucoseInput: Double?
+    public var insulinSensitivity: Double?
+    public var carbohydrateInput: Double?
+    public var insulinCarbohydrateRatio: Double?
+    public var bloodGlucoseTarget: BloodGlucoseTarget?
+    public var recommended: Recommended?
+    public var bolus: TBolusDatum?      // NOTE: API write only
+    public var bolusId: String?         // NOTE: API read only
```
nhamming, line 23: Just want to note that I don't understand the process related to the bolus ID. Probably clearer if I understand the backend better. Confusion come from not seeing bolusID in bolus classes, so I assume this is server generated? Then on a read, the bolus ID can be used to pull the specific bolus details? Is there no need to cache the ID locally if the bolus details are updated?

nhamming (approved):
Just questions and nits based on my lack of knowledge. No show stoppers.

TidepoolKit/Model/Datum/BolusDatum/TBolusDatum.swift
darinkrauss: The subclasses can directly access this property after initialization. It is not a very commonly used property so I didn't add it to the subclass initializers (but I added it here for a complete initializer).

TidepoolKit/Model/Datum/TCalculatorDatum.swift
darinkrauss: Name-wise? Yes, I agree. However, this matches the current Tidepool data model. I didn't want to get too much into renaming properties different from the Tidepool data model. (I did expand some abbreviations for some properties in some types. For example, `bgTarget` expanded to `bloodGlucoseTarget`. These expansions are already marked as `TODO`s on the backend platform code so I figured it was reasonable.)

TidepoolKit/Model/Datum/TCalculatorDatum.swift
darinkrauss: Yea, it is odd. The Tidepool data model documentation describes how this all works. 

Honestly, this is better than how the actual API works. The API (aka Tidepool data model) uses the *same* field name ('bolus') for *both* the bolus datum (a `struct` on input) and the bolus id (a `string` on output). This was included in the very first data model before my time. This is backend tech debt that has not yet been addressed.

PR #15, opened 2020-04-07
Various updates to support actual data from Tidepool Service and backend updates

darinkrauss:
- https://tidepool.atlassian.net/browse/LOOP-340
- Rename dosing decision bloodGlucoseTargetRangeSchedule to bloodGlucoseTargetSchedule
- Add dosing decision bloodGlucoseForecastIncludingPendingInsulin
- Add time to dosing decision recommended basal and bolus
- Change dosing decision recommended basal duration to seconds in one day
- Rename pump settings dosingEnabled to automatedDelivery
- Use common pump settings schedule time zone offset
- Add pump settings bloodGlucoseTargetPhysicalActivity
- Remove separate pump settings suspend threshold struct
- Add separate pump settings insulin model struct
- Add insulin units to pump settings


rickpasetto (approved):
Again, I can't vouch as to the accuracy of this vis a vis the Tidepool API spec.  But the code looks 🆗

TidepoolKit/Model/Datum/TPumpSettingsDatum.swift
```diff
@@ -286,6 +268,28 @@ public class TPumpSettingsDatum: TDatum, Decodable {
...
+            case fiasp
+            case other
+            case rapidAdult
+            case rapidChild
+            case walsh
+        }
+
+        public var modelType: ModelType?
+        public var modelTypeOther: String?
+        public var actionDuration: Int?
```
ps2, line 283: `actionDuration` and `actionPeakOffset` are parameters to a generalized model we call the "exponential" insulin model. Currently, walsh is the only model supported by Loop that has a configurable parameter (duration).  Is it possible to model like this:
```
public enum ModelType: String, Codable {
    case fiasp
    case rapidAdult
    case rapidChild
    case walsh(duration: TimeInterval)
}
```
"Other" seems like a premature generalization, or is it to support some other feature I'm not aware of?

ps2 (commented):
I think the `InsulinModel` might need some changes; added comments.

TidepoolKit/Model/Datum/TPumpSettingsDatum.swift
darinkrauss: The `actionDuration` is just `effectDuration` in the `LoopKit` `InsulinModel` protocol. It has a different name in the Tidepool data model to be more consistent with the rest of the Tidepool data model (where `action` is used in relation to insulin action).

The `actionPeakOffset` is more specific to the exponential models, `humalogNovologAdult`, `humalogNovologChild`, and `fiasp`. 

In both of these cases I'm considering the potential for Tidepool Service to be used by an enterprising DIY Looper who has changed any of these values from their defaults. With Tidepool Loop we *know* what these settings will be and so, technically, do not need to include them in the data we send to the backend, but I'm trying to generalize for the DIY case.

This is also why the `other` `modelType` is available, in case someone has created their own insulin model. Yes, I know it is highly unlikely, but I'd rather have a place in the Tidepool data model to support this than having someone be forced to cram their own insulin model into one of the pre-existing enumerations (and it not representing reality). We have this `other` option available elsewhere in the Tidepool data model to cover non-Tidepool products and allow extensibility without requiring an update to the Tidepool data model or forcing a third-party developer to cram data into where it doesn't belong.

I know you are a strong believer in "don't do it unless you have a confirmed, supported use case for it" and while we definitely don't have a use case within Tidepool Loop, I do believe the Tidepool data model does need to support third-party developers uploading their own data to the backend.

darinkrauss: Oh, and regarding the proposed `enum`, yes that it much more Swift-like, but it is not Tidepool data model-like. I'm trying to thread the fine line between Swift support and being as similar to the Tidepool data model as possible. Supporting `enum`s like this is possible, but it makes it more difficult to follow back to the Tidepool data model and also requires additional `Codable` code to correctly map from the Swift data type into the JSON required for the Tidepool data model.

ps2: If you want to support arbitrary models based on the Exponential model, then exponential should be a type, with its parameters specified. 
If we have a enum value for "rapidAdult", that should mean something. If someone specifies "rapidAdult", and doesn't use the rapidAdult parameters what is that? It's not the rapidAdult model. 
I think if you want to support arbitrary insulin models, you'd want to support a name plus a parameter dictionary.

ps2: When considering more "swift-like" modeling, it's important to remember that the primary mission of this layer is to convert the json-y data model into a type system that swift apps can use easily. The type system can let you enforce the data model more precisely, preventing more user errors.  I understand it's a tradeoff, and I'm guessing you are considering these things. I feel that the Swift type system offers some really helpful tools here, but understand if you are making a pragmatic tradeoff.

darinkrauss: Yes, that makes sense. How about we leave this as-is for now and as I continue working with `TidepoolKit` to implement `TidepoolService` I evolve this modeling to be more "swift-life" where it makes sense? 

For example, per an earlier suggestion to use `TimeInterval` (rather than an `Int`) to describe time durations, I think I've convinced myself that using `TimeInterval` is worth the extra code (and potential human error in any code) to offset the confusion of the various `Int` durations (where some are in minutes, some seconds, and some milliseconds). I added a ticket for this specific work last Monday, https://tidepool.atlassian.net/browse/LOOP-1127.

How does that sound?

ps2 (approved):

PR #16, opened 2020-04-21
LOOP-340 Integrate latest TidepoolKit into TidepoolService (part 2)

darinkrauss:
- https://tidepool.atlassian.net/browse/LOOP-340
- Add completion handler to delegate login signup callback
- Update sample dosing decision to use negative insulin on board
- Fix crashing bug only in release optimized code with DNS query callback


nhamming (approved):
LGTM

rickpasetto (approved):
Nice 👍 LGTM

PR #18, opened 2021-04-19
[LOOP-3375] Onboarding service integration

darinkrauss:
- https://tidepool.atlassian.net/browse/LOOP-3375
- Replace login UI storyboard with SwiftUI implementation
- Add proper cancel workflow
- Rename TLoginSignup delegate to loginSignupDelegate to avoid naming conflict
- Remove unnecessary assets and storyboard
- Add various support and extension code
- Set minimum deployment target to iOS 14.0


TidepoolKitUI/Internal/Previews/ContentPreview.swift
```diff
@@ -0,0 +1,61 @@
...
+            content
+                .colorScheme(.dark)
+                .environment(\.sizeCategory, .accessibilityExtraExtraExtraLarge)
+                .previewDevice(PreviewDevice(rawValue: "iPhone SE (2nd generation)"))
+                .previewDisplayName("iPhone SE (2nd generation) - Dark - AXXXL")
+            content
+                .colorScheme(.light)
+                .environment(\.sizeCategory, .extraSmall)
+                .previewDevice(PreviewDevice(rawValue: "iPhone 12 Pro Max"))
+                .previewDisplayName("iPhone 12 Pro Max - Light - XS")
```
nhamming, line 37: I like this. What about adding an iPod size?

nhamming (approved):
LGTM

TidepoolKitUI/Internal/Previews/ContentPreview.swift
darinkrauss: I would replacing one of the above with an iPod (in the future), but not adding any more. (Each additional one has a resource hit and takes more time to render.) What is our minimum resolution device?

PR #19, opened 2021-05-07
[LOOP-1274] Fetch prescription from Tidepool backend

darinkrauss:
- https://tidepool.atlassian.net/browse/LOOP-1274
- Add Prescription and PrescriptionClaim types
- Add TAPI.claimPrescription endpoint
- Move session maintenance internally within TAPI
- Automatically refresh session after expiration and authentication failures
- Add TAPIObserver to allow monitoring of session updates
- Add TSession created date to determine if automatic refresh is desired
- Add, updated, and refactor tests and mocks

### DIFF
--- a/TidepoolKit/Internal/WeakSynchronizedSet.swift
+++ b/TidepoolKit/Internal/WeakSynchronizedSet.swift
@@ -1,0 +1,147 @@
+//
+//  WeakSynchronizedSet.swift
+//  TidepoolKit
+//
+//  Created by Michael Pangburn
+//  Copyright © 2019 LoopKit Authors. All rights reserved.
+//
+
+import Foundation
+
+
+/// A set-like collection of weak types, providing closure-based iteration on a client-specified queue
+/// Mutations and iterations are thread-safe
+public class WeakSynchronizedSet<Element> {
+    private typealias Identifier = ObjectIdentifier
+    private typealias ElementContainer = ElementDispatchContainer<Element>
+
+    private class ElementDispatchContainer<Element> {
+        private weak var _element: AnyObject?
+        weak var queue: DispatchQueue?
+
+        var element: Element? {
+            return _element as? Element
+        }
+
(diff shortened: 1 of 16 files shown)


darinkrauss:
LW PR for build: https://github.com/tidepool-org/LoopWorkspace/pull/516

TidepoolKit/Internal/WeakSynchronizedSet.swift
```diff
@@ -0,0 +1,147 @@
...
+//  Created by Michael Pangburn
+//  Copyright © 2019 LoopKit Authors. All rights reserved.
+//
+
+import Foundation
+
+
+/// A set-like collection of weak types, providing closure-based iteration on a client-specified queue
+/// Mutations and iterations are thread-safe
+public class WeakSynchronizedSet<Element> {
```
rickpasetto: Hm...maybe not make this `public`?  We don't want name collisions...

TidepoolKit/Model/Prescription/TPrescription.swift
```diff
@@ -0,0 +1,232 @@
+//
+//  TPrescription.swift
+//  TidepoolKit
+//
+//  Created by Darin Krauss on 4/22/21.
+//  Copyright © 2021 Tidepool Project. All rights reserved.
+//
+
+public struct TPrescription: Codable, Equatable {
```
rickpasetto, line 9: FWIW I don't know the official data model, so this review doesn't include my checking on that.  If you want me to re-review it with that in mind, LMK.

rickpasetto (approved):
Some comments, one suggestion (please consider it), none really blocking.  LGTM.

TidepoolKit/Internal/WeakSynchronizedSet.swift
darinkrauss: Ah, good point.

PR #21, opened 2021-06-25
[LOOP-3611] Prevent use if the user has a jailbroken iOS device

darinkrauss:
- https://tidepool.atlassian.net/browse/LOOP-3611
- Add verify device API
- Add tests


rickpasetto (approved):
LGTM

PR #23, opened 2021-08-03
[LOOP-3791] addition logging in device check

nhamming:

code reviewed with #22

### DIFF
--- a/TidepoolKit/TAPI.swift
+++ b/TidepoolKit/TAPI.swift
@@ -387,8 +387,10 @@ public class TAPI {
         performRequest(request) { (result: DecodableResult<VerifyDeviceResponseBody>) -> Void in
             switch result {
             case .failure(let error):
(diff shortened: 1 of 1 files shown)


TidepoolKit/TAPI.swift
```diff
@@ -387,8 +387,10 @@ public class TAPI {
         performRequest(request) { (result: DecodableResult<VerifyDeviceResponseBody>) -> Void in
             switch result {
             case .failure(let error):
+                TSharedLogging.debug("verify device request failed. deviceToken:\(deviceToken), error:\(error)")
```
rickpasetto: Just a quick [nit]: this will output the actual base64 `deviceToken`, right?  Otherwise we can't use it for testing later...

nhamming: ah... good catch. I'll update in `release`

rickpasetto (approved):
LGTM

PR #24, opened 2021-08-04
[LOOP-3791] log the encoded device token

nhamming:


rickpasetto (approved):
Yeah, I was afraid of that.  Thanks...
LGTM

PR #25, opened 2021-08-05
[LOOP-3791] instantiate TLSharingLogging and log all device check paths

nhamming:


rickpasetto (approved):
LGTM

PR #29, opened 2021-09-07
LOOP-1114: Adds `TInfo` type to TidepoolKit

rickpasetto:
This is in support of the API to get Tidepool version info.  See https://tidepool.atlassian.net/browse/LOOP-1114?focusedCommentId=36922 for the schema of the API response returned.

https://tidepool.atlassian.net/browse/LOOP-1114


TidepoolKit/Model/TInfo.swift
```diff
@@ -0,0 +1,79 @@
+//
+//  TInfo.swift
+//  TidepoolKit
+//
+//  Created by Rick Pasetto on 9/7/21.
+//  Copyright © 2021 Tidepool Project. All rights reserved.
+//
+
+public struct TInfo: Codable, Equatable {
```
darinkrauss, line 9: At this level (which isn't supposed to know anything about Loop or necessarily what any of the fields mean or do not mean; this library is agnostic), I think this should be trimmed down to:
```
public struct TInfo: Codable, Equatable {
    public var versions: Versions?

    public struct Versions: Codable, Equatable {
        public var uploaderMinimum: String?
        public var loop: Loop?

        public struct Loop: Codable, Equatable {
            public var minimumSupported: String?
            public var criticalUpdateNeeded: [String]?
        }
    }
}
```
The rest of the logic that is specific to Loop should move into LoopKit (not TidepoolService, which is also supposed to be generic; though there may be a translation layer there from the TidepoolKit type (which includes a bunch of unnecessary stuff like uploaderMinimum) to a LoopKit type which would just have the Loop version info and none of the wrappers) which will call this API.

BTW, structs within other structs do not use the "T" prefix.

TidepoolKit/TAPI.swift
```diff
@@ -206,6 +206,19 @@ public class TAPI {
         }
     }
 
+    // MARK: - Info
+
+    /// Get the supported version info for apps.
```
darinkrauss: [nit] Just "Get the info." because it may be expanded to include much more than version info.

TidepoolKit/TAPI.swift
```diff
@@ -206,6 +206,19 @@ public class TAPI {
...
+    ///
+    /// - Parameters:
+    ///   - environment: The environment to check version info for.
```
darinkrauss: [nit] "The environment to get the info for."

darinkrauss (changes requested):
I think a bunch of this needs to move "up" to a higher level. See comments. Let me know if you have any questions/comments.

TidepoolKit/Model/TInfo.swift
rickpasetto: I hear what you're saying, but, there's a "Loop" right there in the struct...and the nature of the API itself is to talk about app versions.  Not sure you can avoid "loop-ness".  But I'll make the change you suggest.

TidepoolKit/TAPI.swift
rickpasetto: Admittedly it's pretty dodgy to have just an `/info` endpoint without any context.  Seems like it could eventually be a dumping ground... 🤷

darinkrauss: Yep, but its been around for a long while so I doubt it will change.

TidepoolKit/TAPI.swift
```diff
@@ -206,6 +206,19 @@ public class TAPI {
...
+    // MARK: - Info
+
+    /// Get the info.
+    ///
+    /// - Parameters:
+    ///   - environment: The environment to get the info for.
+    ///   - completion: The completion function to invoke with any error.
+    public func getInfo(environment: TEnvironment? = nil, completion: @escaping (Result<TInfo, TError>) -> Void) {
+        // Note: no session is needed
+        let request = createRequest(environment: environment ?? environments.first!, method: "GET", path: "/info")
```
darinkrauss: You'll need to include the current session environment (if one), before defaulting to the `environments.first!`. So,
```
let request = createRequest(environment: environment ?? session?.environment ?? environments.first!, method: "GET", path: "/info")
```

darinkrauss (changes requested):
Once change required to included the active session environment before the default environment.

darinkrauss (approved):
LGTM!

darinkrauss (approved):
LGTM! Thanks for making those changes.

PR #31, opened 2021-10-16
LOOP-1114: Enable configuration of the default Tidepool Environment using appGroup

rickpasetto:
This makes it possible to build TidepoolKit in multiple apps within an appGroup (denoted by the special Info.plist key, "AppGroupIdentifier"), and have one app configure the default TEnvironment TAPI uses.

### DIFF
--- a/TidepoolKit/Internal/Extensions/UserDefaults.swift
+++ b/TidepoolKit/Internal/Extensions/UserDefaults.swift
@@ -26,7 +26,7 @@ extension UserDefaults {
             return try? JSONDecoder.tidepool.decode(TEnvironment.self, from: data)
         }
         set {
-            set(try! JSONEncoder.tidepool.encode(newValue), forKey: Key.defaultEnvironment.rawValue)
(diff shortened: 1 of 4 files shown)


TidepoolKit/Internal/Extensions/UserDefaults.swift
```diff
@@ -0,0 +1,37 @@
...
+    
+    var defaultEnvironment: TEnvironment? {
+        get {
+            guard let data = data(forKey: Key.defaultEnvironment.rawValue) else {
+                return nil
+            }
+            return try? JSONDecoder.tidepool.decode(TEnvironment.self, from: data)
+        }
+        set {
+            set(try! JSONEncoder.tidepool.encode(newValue), forKey: Key.defaultEnvironment.rawValue)
```
darinkrauss: [nit] `try!`?

darinkrauss (approved):
Just one nit, but LGTM!

PR #33, opened 2021-11-11
[LOOP-28] Upload loop settings data to Tidepool backend

darinkrauss:
- https://tidepool.atlassian.net/browse/LOOP-28
- Add TControllerSettingsDatum
- Add TPumpSettingsOverrideDeviceEventDatum
- Add fields to TCGMSettingsDatum
- Update fields of TPumpSettingsDatum
- Remove TApplicationSettingsDatum as unnecessary
- Add insulin formulation to bolus types
- Add tests


darinkrauss:
LW PR for build: https://github.com/tidepool-org/LoopWorkspace/pull/698

TidepoolKit/TLogging.swift
```diff
@@ -42,7 +42,7 @@ public protocol TLogging {
 public class TSharedLogging {
 
     /// A shared, global instance of TLogging.
-    public static var instance: TLogging?
+    public static weak var instance: TLogging?
```
rickpasetto: Who is supposed to set this?  Seems like an odd contract to have a singleton that doesn't hold a ref to this.  Now, anyone setting this like this:
```
class MyLogging: TLogging { ... }
TSharedLogging.instance = MyLogging()
```

will wind up losing their Logger.

rickpasetto (commented):
Question about the contract regarding who should own the `TLogging` singleton.

TidepoolKit/TLogging.swift
darinkrauss: Fixed. Logger is now a simple property on the `TAPI` object.

rickpasetto (approved):
👍

PR #38, opened 2022-02-23
COASTAL-460: Adds triggerDateMatching for `.nextDate` and `.nextDateRepeating`

rickpasetto:
Note: DateComponents already conforms to Codable

https://tidepool.atlassian.net/browse/COASTAL-460

### DIFF
--- a/TidepoolKit/Model/Datum/TAlertDatum.swift
+++ b/TidepoolKit/Model/Datum/TAlertDatum.swift
@@ -30,10 +30,10 @@ public class TAlertDatum: TDatum, Decodable {
     }
     
     public struct TriggerNextDateTimeSpec: Codable, Equatable {
-        public var dayOfMonth: Int? // 1-31
-        public var hourOfDay: Int? // 0-23
-        public var minuteOfHour: Int? // 0-59
-        public init(dayOfMonth: Int? = nil, hourOfDay: Int? = nil, minuteOfHour: Int? = nil) {
+        public var dayOfMonth: Int // 1-31
+        public var hourOfDay: Int // 0-23
+        public var minuteOfHour: Int // 0-59
+        public init(dayOfMonth: Int, hourOfDay: Int, minuteOfHour: Int) {
             self.dayOfMonth = dayOfMonth
             self.hourOfDay = hourOfDay
             self.minuteOfHour = minuteOfHour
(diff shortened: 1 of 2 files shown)


TidepoolKit/Model/Datum/TAlertDatum.swift
```diff
@@ -25,12 +25,15 @@ public class TAlertDatum: TDatum, Decodable {
...
         case repeating
+        case nextDate
+        case nextDateRepeating
     }
     
     public var name: String?
     public var priority: Priority?
     public var trigger: Trigger?
     public var triggerDelay: TimeInterval?
+    public var triggerDateMatching: DateComponents?
```
darinkrauss: I think we're going to need to create a TidepoolKit type that partially mirrors `DateComponents`. The reason for doing this is that the backend will need to be updated to support this new field and data type. However, I don't think it is reasonable to expect that the backend support every known option of `DateComponents` as there are a lot that we won't likely use (e.g. era, yearForWeekOfYear, isLeapMonth, ...). I think we need to explicitly state we support only certain fields and define a type here to capture that.

TidepoolKitTests/Model/Datum/TAlertDatumTests.swift
```diff
@@ -10,11 +10,13 @@ import XCTest
 import TidepoolKit
 
 class TAlertDatumTests: XCTestCase {
+    static let noon = DateComponents(hour: 12, minute: 0)
```
darinkrauss: [nit] We should test something that includes all of the *supported* date components. (See previous comment.)

darinkrauss (changes requested):
See comments. Plus, what is the plan for updating the backend to support these changes? If we merge this now it will entirely break uploading alerts and I don't think we want that regression.

TidepoolKit/Model/Datum/TAlertDatum.swift
rickpasetto: @darinkrauss and I chatted and I've agreed to make a specific type that only supports `days`, `hours`, and `minutes` for both `LoopKit` and `TidepoolKit`.  Stay tuned for the changes...

rickpasetto:
After discussion with @ps2, product and partner, decided to not go forward with this change.

PR #41, opened 2023-01-26
[COASTAL-1108] App attestation

nhamming:
https://tidepool.atlassian.net/browse/COASTAL-1108


TidepoolKit/TAPI.swift
```diff
@@ -432,6 +432,105 @@ public class TAPI {
         let valid: Bool
     }
 
+    // MARK: - Verify App
```
ps2, line 435: nit: This looks pretty self contained (no new vars, just functions and types), and this file is pretty big. Maybe it'd be nice as an extension?

ps2 (approved):
LGTM! Just a small non-blocking suggestion.

PR #47, opened 2023-04-03
[LOOP-4483] Study Product Selection

Camji55:
https://tidepool.atlassian.net/browse/LOOP-4483


nhamming (commented):
hm... not sure why this needs to reside in this repo. I think the name of the repo is misleading. I believe this repo is intended to only interface with the Tidepool Backend. Maybe it should be named as such (TidepoolBackendKit or similar).

PR #48, opened 2023-04-03
[LOOP-4483] Study Product Selection

Camji55:


Sources/TidepoolKit/Internal/Extensions/UserDefaults.swift
```diff
@@ -29,6 +30,15 @@ extension UserDefaults {
...
     }
+    
+    public var studyProductSelection: String? {
+        get {
+            string(forKey: Key.studyProductSelection.rawValue)
+        }
+        set {
+            set(newValue, forKey: Key.studyProductSelection.rawValue)
+        }
+    }
```
nhamming, line 41: I believe this repo is just the interface to Tidepool Backend. I think the repo name is confusing, since it is not a general kit for Tidepool and is made available as a `Common` plugin. Probably should be named `TidepoolBackendKit` or similar. 

Since this is in `Common`, I think this code is better placed elsewhere. I'm starting to understand better why you originally wanted it in LoopKit, since you need it in a common repo. Wonder if we should chat about this?

PR #49, opened 2023-04-14
[LOOP-4596] Scenario Organization

Camji55:
https://tidepool.atlassian.net/browse/LOOP-4596

### DIFF
--- a/Sources/TidepoolKit/Internal/Extensions/UserDefaults.swift
+++ b/Sources/TidepoolKit/Internal/Extensions/UserDefaults.swift
@@ -16,7 +16,6 @@ extension UserDefaults {
 
     private enum Key: String {
         case defaultEnvironment = "org.tidepool.TidepoolKit.DefaultEnvironment"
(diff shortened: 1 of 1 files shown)


Sources/TidepoolKit/Internal/Extensions/UserDefaults.swift
```diff
@@ -16,6 +16,7 @@ extension UserDefaults {
 
     private enum Key: String {
         case defaultEnvironment = "org.tidepool.TidepoolKit.DefaultEnvironment"
+        case studyProductSelection = "org.tidepool.plugins.TidepoolSupport.StudyProductSelection"
```
ps2: This is outside TidepoolKit's responsibilities; this is a general purpose library for interacting with the Tidepool SDK, and these are internal Tidepool Loop concerns. Would probably be better in the TidepoolSupport framework, and the Tidepool Loop Configurator could access this code from there.

ps2 (changes requested):

PR #50, opened 2023-04-21
LOOP-887 Updates for KeyCloak

ps2:
https://tidepool.atlassian.net/browse/LOOP-887

Incorporates AppAuth-iOS to help perform OAuth2 based authorization, instead of the old login/password authentication method.
Also updates to Swift async to reduce nesting of callbacks


ExampleApp/TidepoolKit Example/LoginView.swift
```diff
@@ -0,0 +1,191 @@
...
+//  TidepoolKit Example
+//
+//  Created by Pete Schwamb on 4/13/23.
+//  Copyright © 2023 Tidepool Project. All rights reserved.
+//
+
+import SwiftUI
+import TidepoolKit
+
+@MainActor
```
Camji55, line 12: `@MainActor` here should not be needed

Sources/TidepoolKit/Auth/Authorization.swift
```diff
@@ -0,0 +1,118 @@
...
+    }
+}
+
+class AppAuthAuthorization: Authorization {
+    var currentAuthorizationFlow: OIDExternalUserAgentSession?
+
+    func presentAuth(request: OIDAuthorizationRequest, presenting: UIViewController) async throws -> AuthorizationState {
+        // Present the authentication session using AppAuth
+        let result: (OIDAuthState, OIDExternalUserAgentSession) = try await withCheckedThrowingContinuation { continuation in
+            DispatchQueue.main.async {
```
Camji55: To stick within the world of new the concurrency model, you can mark this function as `@MainActor` or `await MainActor.run { ...`

Camji55: Is main queue even needed here?

ExampleApp/TidepoolKit Example/LoginView.swift
ps2: The `body` method of SwiftUI views is automatically annotated with `@MainActor`, but if I want to ensure all other methods/accesses happen on the main actor, then I need to do this, or annotate the individual methods, correct? In this specific case, I'm talking about things like the login/logout button actions. Let me know if there is a preferred style for this.

Sources/TidepoolKit/Auth/Authorization.swift
ps2: TBH, I really want to drop AppAuth and just use ASWebAuthenticationSession directly; we're not getting much from AppAuth at this point. But yes, we do need to call this from the main thread, I think. The docs don't specify, but it presents a view controller, and tracing through the code, I don't see it ever switching over to main inside the library for this call.

Sources/TidepoolKit/Model/TUser.swift
```diff
@@ -0,0 +1,33 @@
...
+    "name@email.org"
+ ],
+ "roles":[
+    "default-roles-integration",
+    "patient"
+ ],
+ "termsAccepted":"2023-01-19T21:34:27+00:00",
+ "userid":"e631fc5a-1234-4686-a498-8f2bbfec55b6",
+ "username":"name@email.org"
+} */
```
nhamming: If this example should be kept, consider including a comment explaining it.

Sources/TidepoolKit/TAPI.swift
```diff
@@ -39,43 +36,57 @@ public class TAPI {
     /// The URLSessionConfiguration used for all requests. The default is typically acceptable for most purposes. Any changes
     /// will only apply to subsequent requests.
     public var urlSessionConfiguration: URLSessionConfiguration {
```
nhamming: Why is this public when there is a `setURLSessionConfiguration` below? Is this redundancy needed?

Sources/TidepoolKit/TAPI.swift
```diff
@@ -39,43 +36,57 @@ public class TAPI {
...
     /// The session used for all requests.
     public var session: TSession? {
-        get {
-            return sessionLocked.value
-        }
-        set {
-            sessionLocked.mutate { $0 = newValue }
-            observers.forEach { $0.apiDidUpdateSession(newValue) }
+        didSet {
+            observers.forEach { $0.apiDidUpdateSession(self.session) }
```
nhamming: just wanted to confirm that the removal of locked parameters is result of using `@MainActor`, right?

Sources/TidepoolKit/TAPI.swift
nhamming: similar comment for other parameters (e.g., `session`).

Sources/TidepoolKit/TAPI.swift
```diff
@@ -95,71 +106,134 @@ public class TAPI {
...
+
+        guard let response = response as? HTTPURLResponse else {
+            throw TError.responseUnexpected(response, data)
+        }
+
+        guard response.statusCode >= 200 && response.statusCode <= 299 else {
+            throw TError.responseUnexpectedStatusCode(response, data)
+        }
+
+        // Token has successfully been
```
nhamming: been what? Is this implementation incomplete or just the comment?

Sources/TidepoolKit/TAPI.swift
```diff
@@ -95,71 +106,134 @@ public class TAPI {
...
+
+    /// Login to the Tidepool environment using AppAuth (OAuth2/OpenID-Connect)
     /// used internally by the LoginSignupViewController.
     ///
     /// - Parameters:
     ///   - environment: The environment to login.
-    ///   - email: The email to use for login.
-    ///   - password: The password to use for login.
+    ///   - presenting: A UIViewController to present the login modal from. Can be UIApplication.shared.windows.first!.rootViewController!
     ///   - completion: The completion function to invoke with any error.
```
nhamming: there is no longer a completion.

Sources/TidepoolKit/TAPI.swift
```diff
@@ -238,14 +314,13 @@ public class TAPI {
     /// - Parameters:
     ///   - userId: The user id for which to get the profile. If no user id is specified, then the session user id is used.
     ///   - completion: The completion function to invoke with any error.
-    public func getProfile(userId: String? = nil, completion: @escaping (Result<TProfile, TError>) -> Void) {
+    public func getProfile(userId: String? = nil) async throws -> TProfile {
```
nhamming: [general comment] it seem comments for a few functions have not be updated to align with the changes. Please review all and update.

nhamming (approved):
I think some clean-up is still needed. But I didn't detect any issues.

Have tests been updated to include these changes? Is there good test coverage? If either of these are not true, considering addressing now.

Sources/TidepoolKit/TAPI.swift
ps2: Revoked. :) Will update. Thanks.

Sources/TidepoolKit/TAPI.swift
ps2: Yes

Sources/TidepoolKit/Model/TUser.swift
ps2: Will remove

Sources/TidepoolKit/TAPI.swift
ps2: Good observation. Will try to resolve this redundancy.

ps2:
@nhamming @Camji55 Updates from your reviews have been pushed. Thanks!  I also added some additional code (See Auth/OAuth2Authenticator.swift) to allow an OAuth2 flow without the use of ViewControllers (Care Partner app is all SwiftUI).  This new flow also paves the way to removing the (old and barely maintained) objective-c AppAuth framework, once TidepoolService and the example app are updated to use it.

Camji55 (approved):

Sources/TidepoolKit/TAPI.swift
ps2: Well, it's a result of making this class an actor; it runs on its own space, not main.

nhamming (approved):
LGTM

PR #51, opened 2023-05-12
CPA-92 Add api for creating invitation

ps2:
https://tidepool.atlassian.net/browse/CPA-92

### DIFF
--- a/Sources/TidepoolKit/Model/TInvite.swift
+++ b/Sources/TidepoolKit/Model/TInvite.swift
@@ -1,0 +1,46 @@
+//
+//  TInvite.swift
+//  TidepoolKit
+//
+//  Created by Pete Schwamb on 5/11/23.
+//
+
+import Foundation
+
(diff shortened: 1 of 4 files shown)


Sources/TidepoolKit/Model/TInvite.swift
```diff
@@ -0,0 +1,46 @@
...
+    }
+}
+
+public struct TInvite: Codable {
+    let key: String
+    let type: String
+    let email: String
+    let creatorId: String
+    let created: Date
+    let modified: Date
```
Camji55, line 42: Nit: maybe these can be named `createdAt` and `modifiedAt` to be a bit more explicit?

Camji55 (approved):

Sources/TidepoolKit/Model/TInvite.swift
ps2: We match the names defined in the API.

PR #52, opened 2023-05-21
Remove AppAuth dependency

ps2:
Same functionality.  A lot less code.

Also includes some fixes to match Tidepool API attribute names: `types` -> `type`, and `subTypes` -> `subType`


Sources/TidepoolKit/Auth/OAuth2Authenticator.swift
```diff
@@ -9,19 +9,53 @@
...
+                    continuation.resume(throwing: TError.missingAuthenticationState)
+                    return
+                }
+
+                continuation.resume(returning: callbackURL)
+            }
+            authenticationSession?.presentationContextProvider = contextProviding
+            authenticationSession?.start()
+        }
+        return callbackURL
```
Camji55: Did you need to declare the variable and type here to satisfy the compiler? I know sometimes it's finicky with the `continuations`. I think you might be able to get around that with this:
```swift
public func startSession(authURL: URL, callbackScheme: String?) async throws -> URL {
    try await withCheckedThrowingContinuation<URL> { continuation in
        ...
    }
}
```

Sources/TidepoolKit/TAPI.swift
```diff
@@ -296,35 +232,47 @@ public actor TAPI {
 
         let config = try await lookupOIDConfiguration(environment: session.environment)
 
-        guard let oidConfig = OIDServiceConfiguration(from: config) else {
+        var components = URLComponents()
+        components.queryItems = [
+            URLQueryItem(name:"refresh_token", value: refreshToken),
+            URLQueryItem(name:"grant_type", value: "refresh_token"),
```
Camji55, line 238: Just checking this is correct?

Camji55 (approved):

Sources/TidepoolKit/TAPI.swift
ps2: Are you asking about the `grant_type`?  If so, yes, "refresh_token" is the correct grant type when asking to refresh the access token.

Sources/TidepoolKit/Auth/OAuth2Authenticator.swift
ps2: Ah, yeah, thanks. Sometimes the compiler will throw a nonsensical error, and adding explicit types helps it throw a better error.  I fixed the real issue and forgot to cleanup the explicit types. Pushed a change to clean this up.

PR #54, opened 2023-06-05
[CPA-112] Pending Invite Model

ArwainK:
https://tidepool.atlassian.net/browse/CPA-112

API reference:
https://tidepool.stoplight.io/docs/tidepool-api/51fe42362a0fe-get-pending-care-team-invitations

PR #55, opened 2023-06-07
Remove context from TPendingInvite

ArwainK:
Backend returning Dictionary when should be String. This is just a small fix, for context see [#54](https://github.com/tidepool-org/TidepoolKit/pull/54)


ps2 (approved):
I wanted to see what context was defined as in the docs, but they appear to be missing this page now. I think it's fine to remove an unused field for now, though, so approving.

PR #56, opened 2023-06-13
Added method to access cancelInvite API via TidepoolKit

ArwainK:
https://tidepool.atlassian.net/jira/software/c/projects/COASTAL/boards/78?modal=detail&selectedIssue=CPA-113

### DIFF
--- a/Sources/TidepoolKit/TAPI.swift
+++ b/Sources/TidepoolKit/TAPI.swift
@@ -353,6 +353,21 @@ public actor TAPI {
         let request = try createRequest(method: "GET", path: "/confirm/invite/\(userId ?? session.userId)")
         return try await performRequest(request)
     }
+    
+    /// Cancel a pending invite for access to the user account identified by userId. If no user id is specified, then the session user id is used.
+    ///
+    /// - Parameters:
+    ///   - userId: The user id. If no user id is specified, then the session user id is used.
+    ///   - invitedByEmail: The email that has been invited.
+    /// - Returns: A confirmation/response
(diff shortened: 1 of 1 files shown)


Sources/TidepoolKit/TAPI.swift
```diff
@@ -353,6 +353,21 @@ public actor TAPI {
...
+    /// - Parameters:
+    ///   - userId: The user id. If no user id is specified, then the session user id is used.
+    ///   - invitedByEmail: The email that has been invited.
+    /// - Returns: A confirmation/response
+    public func cancelInvite(userId: String? = nil, invitedByEmail: String) async throws -> String {
+        guard let session = session else {
+            throw TError.sessionMissing
+        }
+
+        let request = try createRequest(method: "PUT", path: "/confirm/\(userId ?? session.userId)/invited/\(invitedByEmail)")
```
ps2, line 368: A "cancel" is implemented by a "PUT" to `/confirm/...`?  That seems odd to me. I don't see anything in the http side of this that indicates a "cancel". Wish we had the api site up.

ArwainK: Agree, but possibly idempotent behavior is desirable in case of multiple cancel calls? Not sure about the placement in `/confirm/...` Tracking this API: 
https://github.com/tidepool-org/TidepoolApi/blob/master/reference/confirm.v1.yaml#L372-L393 

Tested the call on a few accounts, works to remove pending invites but not accepted invites with `Sharing` status.

ps2 (approved):
LGTM!

PR #60, opened 2023-06-16
[CPA-110] added reject pending invite

nhamming:
https://tidepool.atlassian.net/browse/CPA-110

Also corrects accept pending invite

### DIFF
--- a/Sources/TidepoolKit/TAPI.swift
+++ b/Sources/TidepoolKit/TAPI.swift
@@ -388,16 +388,34 @@ public actor TAPI {
     /// - Parameters:
     ///   - userId: The user id. If no user id is specified, then the session user id is used.
     ///   - invitedByUserId: The user id of the user that sent the invited.
-    /// - Returns: A confirmation/response
-    public func acceptInvite(userId: String? = nil, invitedByUserId: String) async throws -> String {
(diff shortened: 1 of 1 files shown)


Sources/TidepoolKit/TAPI.swift
```diff
@@ -388,16 +388,34 @@ public actor TAPI {
...
+        guard let session = session else {
+            throw TError.sessionMissing
+        }
+        let body = PendingInviteKeyBody(key: key)
+        let request = try createRequest(method: "PUT", path: "/confirm/dismiss/invite/\(userId ?? session.userId)/\(invitedByUserId)", body: body)
+        try await performRequestNotDecodingResponse(request)
     }
     
+    struct PendingInviteKeyBody: Codable {
+        let key: String
```
ps2, line 416: Is string itself codable?  If so, I think you could just pass `key` as the body.

ps2 (approved):
LGTM.  Just a question about a possible simplification.

Sources/TidepoolKit/TAPI.swift
nhamming: I tried this originally, but it doesn't get the correct formatting.

PR #65, opened 2023-09-12
[COASTAL-1292] adding partner details to app assertion

nhamming:
https://tidepool.atlassian.net/browse/COASTAL-1292


ArwainK (approved):
LGTM! Linking api for ref: https://tidepool.stoplight.io/docs/tidepool-api/branches/BACK-2165-app-attestation/5be0356fac1e4-verify-an-assertion

PR #66, opened 2023-09-30
[CPA-91 & CPA-92] Design Feedback/Updates to Caregiver Management

Camji55:
https://tidepool.atlassian.net/browse/CPA-91 & https://tidepool.atlassian.net/browse/CPA-92


ps2 (approved):
Changes lgtm, assuming the backend has been updated similarly.

Camji55:
> Changes lgtm, assuming the backend has been updated similarly.

@ps2, yes the nickname field in `TPendingInvite` has not been added yet but is being added by the backend. The optional here will allow it to work both now and once it's available. As for `TInvite`, that nickname field is already supported.

PR #69, opened 2024-02-28
[LOOP-4776] Invitation API Updates

Camji55:
https://tidepool.atlassian.net/browse/LOOP-4776

This change is currently only available on the qa4 environment. Awaiting QA testing before appearing on prod.


nhamming (approved):
LGTM

PR #71, opened 2024-06-04
LOOP-1169 Device log uploads

ps2:
https://tidepool.atlassian.net/browse/LOOP-1169

### DIFF
--- a/Sources/TidepoolKit/TAPI.swift
+++ b/Sources/TidepoolKit/TAPI.swift
@@ -764,15 +764,15 @@ public actor TAPI {
     /// List device logs metadata
     ///
     /// - Parameters:
-    ///   - logs: entries to upload
     ///   - start: the start date of the period that contains the log entries
     ///   - end: the end date of the period that contains the log entries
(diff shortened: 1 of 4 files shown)


Camji55 (approved):

Sources/TidepoolKit/TAPI.swift
```diff
@@ -728,6 +729,63 @@ public actor TAPI {
...
+        print("Sending with digest: \(digestStr)")
+        request.setValue(digestStr, forHTTPHeaderField: "Digest")
+
+        return try await performRequest(request)
+    }
+
+    /// List device logs metadata
+    ///
+    /// - Parameters:
+    ///   - logs: entries to upload
```
nhamming: `logs` comment should be removed.

nhamming: also return description should be added.

nhamming (approved):

Sources/TidepoolKit/TAPI.swift
ps2: Thanks!

PR #73, opened 2024-10-16
[LOOP-5065] updated device logs API to allow for null response

nhamming:
https://tidepool.atlassian.net/browse/LOOP-5065

### DIFF
--- a/Sources/TidepoolKit/TAPI.swift
+++ b/Sources/TidepoolKit/TAPI.swift
@@ -766,8 +766,8 @@ public actor TAPI {
     /// - Parameters:
     ///   - start: the start date of the period that contains the log entries
     ///   - end: the end date of the period that contains the log entries
-    /// - Returns: A list of ``TDeviceLogsMetadata`` structures
-    public func listDeviceLogs(start: Date, end: Date) async throws -> [TDeviceLogsMetadata] {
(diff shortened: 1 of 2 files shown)


Sources/TidepoolKit/TAPI.swift
```diff
@@ -766,8 +766,8 @@ public actor TAPI {
     /// - Parameters:
     ///   - start: the start date of the period that contains the log entries
     ///   - end: the end date of the period that contains the log entries
-    /// - Returns: A list of ``TDeviceLogsMetadata`` structures
-    public func listDeviceLogs(start: Date, end: Date) async throws -> [TDeviceLogsMetadata] {
+    /// - Returns: An optional list of ``TDeviceLogsMetadata`` structures. If the list is empty, `nil` is returned
```
ps2: I think this is changing behavior so that any failure to decode json returns a nil.

ps2: After some testing, it looks like the backend is returning "null", which could be considered valid json. 

Nit: I think if we are wanting to return something that means "empty list" from this method, it would be better to return an empty list. Otherwise callers need to wonder what the difference between [] and nil is, and handle both.

ps2 (approved):
LGTM, just a nit.

ps2 (approved):
LGTM!

PR #78, opened 2025-02-24
[COASTAL-1461] Completion handler called twice with delay

nhamming:
https://tidepool.atlassian.net/browse/COASTAL-1461

Unfortunately, I'm not certain why the `ASWebAuthenticationSession` completion handler is being called twice. It seems to only occur when there is a network delay. The callback URL for the second call is as described in crashlytics:

```
... tried to resume its continuation more than once returning org.tidepool.tidepoolkit.auth://redirect?
error=temporarily_unavailable&error_description=authentication_expired&iss=https%3A%2F%2Fauth.qa2.tidepool.org
%2Frealms%2Fqa1!
```

This stops the crash, but obviously does not fix the root issue.


ps2 (approved):
LGTM!

PR #80, opened 2026-06-09
Add podspec file (MOBILE-683)

gniezen:
We're rebooting Tidepool Mobile as [mobile-remix](github.com/tidepool-org/mobile-remix), and since [healthkit-uploader](https://github.com/tidepool-org/healthkit-uploader) is a CocoaPod it requires TidepoolKit to be a CocoaPod too - can we add this podspec file to do this?

Addresses [MOBILE-683](https://tidepool.atlassian.net/browse/MOBILE-683).

### DIFF
--- a/TidepoolKit.podspec
+++ b/TidepoolKit.podspec
@@ -1,0 +1,17 @@
+Pod::Spec.new do |spec|
+  spec.name         = "TidepoolKit"
+  spec.version      = "1.0.0"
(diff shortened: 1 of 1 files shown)


gniezen:
Closing because cocoapods are being deprecated at the end of the year, so reworking healthkit-uploader as a swift package instead, with the necessary workarounds in mobile-remix that expects cocoapods (since React Native still uses cocoapods).

